PR #6, opened 2018-02-04
First Level Derivation feature

Ashish81:
@NicolasDorier  I added Derivativefeature.Direct , now NBXplorer is able to track both 1st level and second level derivations. Please review and suggest for betterment

### DIFF
--- a/NBXplorer/Repository.cs
+++ b/NBXplorer/Repository.cs
@@ -387,8 +387,7 @@ namespace NBXplorer
 			_Engine.Do(tx =>
 			{
 				bool needCommit = false;
-				var featuresPerKeyPaths =
-				new[] { DerivationFeature.Deposit, DerivationFeature.Change }
+				var featuresPerKeyPaths = Enum.GetValues(typeof(DerivationFeature)).Cast<DerivationFeature>()
 				.Select(f => (Feature: f, Path: DerivationStrategyBase.GetKeyPath(f)))
 				.ToDictionary(o => o.Path, o => o.Feature);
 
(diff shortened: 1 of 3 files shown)


NicolasDorier:
Seems good! Can you just fix the indent to tabs?

Ashish81:
It was little difficult Indenting as in VS2017 it shows good, but Github shows incorrect indents.

NBXplorer/Repository.cs
```diff
@@ -589,7 +588,7 @@ private void RefillAvailable(DBreeze.Transactions.Transaction tx, DerivationStra
 					Redeem = derivation.Redeem,
 					DerivationStrategy = strategy,
 					Feature = derivationFeature,
-					KeyPath = DerivationStrategyBase.GetKeyPath(derivationFeature).Derive(index, false)
+					KeyPath = DerivationFeature.Direct == derivationFeature ? new KeyPath(index.ToString()) : DerivationStrategyBase.GetKeyPath(derivationFeature).Derive(index, false)
```
NicolasDorier: why not just 'DerivationStrategyBase.GetKeyPath(derivationFeature).Derive(index, false)' ?

Ashish81: Yes, Correct , actually changed the GetKeyPath() function later.
DerivationStrategyBase.GetKeyPath(derivationFeature).Derive(index, false) == new KeyPath().Derive(1)

NicolasDorier:
@Ashish81 I will put an editorconfig for next time so you don't have this problem.

PR #9, opened 2018-02-06
UTXO returns correct DerivationFeature

Ashish81:
`UTXO.feature` not returning correctly , it always returns the Deposit. It is fixed, please review

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,2 @@
+[*.cs]
+indent_style = tab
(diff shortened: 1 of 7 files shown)


NicolasDorier:
Thanks a lot, but can you instead assign the `DerivationStrategy` in the GetUTXOs method of the controller ? I would like to keep the DTO as dumb as possible.

Ashish81:
Did you mean assign the `DerivationFeature ` in the GetUTXOs?? 

> instead assign the `DerivationStrategy `in the GetUTXOs

NicolasDorier:
Perfect thanks!

Ashish81:
You're welcome

Ashish81:
Wondering how to get UTXOs of an address (or keypath)? One way i guess is to compare the scriptpubkey of the UTXOChanges.  Wont it make sense to add a parameter in GetUTXos method ?? for fast retrieval

PR #10, opened 2018-02-09
docker-compose.regtest.yml fixes

vardars:
I couldn't use environment variables passed from docker-compose because vars didn't include BTC prefixes. Now I can use them with containers. I thought there was a usage change to support multiple coins, but docker-compose file became out-of-date.

### DIFF
--- a/docker-compose.regtest.yml
+++ b/docker-compose.regtest.yml
@@ -11,12 +11,12 @@ services:
       dockerfile: DockerFile
     environment:
       NBXPLORER_NETWORK: regtest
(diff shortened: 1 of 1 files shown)


NicolasDorier:
Good catch thanks!

NicolasDorier:
I willl likely simplify this `docker-compose` later. By sharing the bitcoin datadir as a volume on the two containers, we can remove rpc related config by passing by cookiefile auth.

vardars:
Thank you too my friend, it is really good to share these projects with community. If I can find time, I want to help with these changes too.

PR #30, opened 2018-03-29
Add BCH support for NBXplorer

bazooka70:
I added some support for Bitcoin Cash. 

To run it use:
dotnet run --no-build -c Release -- --network=testnet --datadir="D:\Projects\NBXplorer-BCH\DataDir" --chains=bch

At this point, the NBXplorer server connects to the the Bitcoin ABC Node, and gets new blocks just fine.
Now I need to test the client side. using `client.Track()`, making transactions, and getting notifications from the server.


NicolasDorier:
Nice.

You can check in the tests how to switch implementation between litecoin and bitcoin.
Adapt for BCH, then you can run the tests for bch.

bazooka70:
Something strange I noticed.

In my project I added a "BitcoincashNetworks.cs" which is basically the same as "NBXplorer.Client/Altcoins/Bitcoincash.cs".

e.g.

```
namespace Bitcoincash
{
  public class Networks
  {...
```

Now In my test project I used:

```
Bitcoincash.Networks.EnsureRegistered();
...
var xnetwork = new NBXplorerNetworkProvider(ChainType.Test).GetBCH();
```

The above line throws an exception; 

> An unhandled exception of type 'System.InvalidOperationException' occurred in mscorlib.dll
> The network bch-main is already registered

What could cause this? I had no problem mixing `NBitcoin.Network` with `var xnetwork = new NBXplorerNetworkProvider(ChainType.Test).GetBTC();`

bazooka70:
Actually the same happens with NBitcoin.Litecoin and NBXplorer.Client:
https://github.com/MetacoSA/NBitcoin.Litecoin

e.g.

NBitcoin.Litecoin.Networks.EnsureRegistered();
var network_ltc = NBitcoin.Litecoin.Networks.Mainnet;
var xnetwork_ltc = new NBXplorerNetworkProvider(ChainType.Main).GetLTC(); // <- exception here: The network ltc-main is already registered

NicolasDorier:
no idea, try to debug why it try to register twice.

NicolasDorier:
Good work, I am testing this, it seems to work, outside a small change in ABC address format. I am soon merging this.

NicolasDorier:
I merged it. I had to change the address format (see https://support.bitpay.com/hc/en-us/articles/115004671663-BitPay-s-Adopted-Conventions-for-Bitcoin-Cash-Addresses-URIs-and-Payment-Requests)

Note that I could not completely test it because their RPC is not sending correctly formatted address.

bazooka70:
Thanks! :+1: 
I don't care about the "BCash" naming. it's shorter :)
However I had many thoughts about the the address format. I am going to try to write a convertor similar to:
https://cashaddr.bitcoincash.org/
So I could convert beeten the 3 formats (CashAddr/Copay/Legacy).

I'm not sure if adopting the Bitpay format as **default** is the correct choice (instead of the base58 legacy). 
However, I'm counting on you to know better.
I will continue to investigate further.

NicolasDorier:
On Bitcoin ABC they have a special format which would require refactoring at NBitcoin level, which is out of question.

bazooka70:
I realize that. but even bitpay (and blockchain.info and others) uses the "qr0n5axmtjvrgltcer5ngu4t5mxxzl64egrgqmnxhe" format to recive funds. so there must be a helper class to convert between the different formats.

NicolasDorier:
sure let me know, we can include such helper function inside NBXplorer lib for sure.

bazooka70:
I made a very initial translation from js to c#
I'm not a c# programmer by definition so it might be ugly :-P

Sources:
https://github.com/bitcoincashjs/cashaddrjs/
https://github.com/bitcoincashjs/bchaddrjs/

c#:
https://github.com/bazooka70/BCashAddr/blob/master/BCashAddr.cs

NicolasDorier:
and how to convert a bitpay address to a BchAddr ?

bazooka70:
```
BchAddrData bchAddr = BchAddr.DecodeAddress("CTH8H8Zj6DSnXFBKQeDG28ogAS92iS16Bp");
var cashaddr = bchAddr.AsCashaddrAddress; // bitcoincash:qpm2qsznhks23z7629mms6s4cwef74vcwvy22gdx6a
var cashaddrNoPrefix = bchAddr.AsCashaddrAddressNoPrefix; // qpm2qsznhks23z7629mms6s4cwef74vcwvy22gdx6a
```

NicolasDorier:
Nice,

PR #33, opened 2018-04-01
Support Polis (mainnet and testnet) - Ready to merge

cronos-polis:
Don't merge yet, we are running tests.


cronos-polis:
Working correctly. I need to downgrade protocol version, it was using bitcoin 70012 and we are on 70208. Is there a way to override variable from NBitcoin?

NicolasDorier:
hey, can you add POLIS to https://github.com/MetacoSA/NBitcoin.TestFramework/blob/master/NBitcoin.TestFramework/WellknownNodeDownloadData.cs ? This would allow me to run unit tests against it.

NicolasDorier:
You can pass a parameter here: 

https://github.com/dgarage/NBXplorer/blob/0917fe66cb84883cc2d740bae04430dc7a7fc776/NBXplorer/BitcoinDWaiter.cs#L363 of type `NodeRequirement`. Can you try using this?

cronos-polis:
Thanks for pointing, I will try to use it to pass the protocol.
I am currently compiling new wallets, Linux is uploading, once I finish I'll update on NBitcoin.TestFramework.

cronos-polis:
Here is, just linux working now, I'm uploading Windows and OSX:
https://github.com/polispay/NBitcoin.TestFramework

NicolasDorier:
Once you are done, I will make a new version of the TestFramework, then you can try to run the tests.

NBXplorer.Client/NBXplorerNetworkProvider.cs
```diff
@@ -11,7 +11,8 @@ public NBXplorerNetworkProvider(ChainType chainType)
 			InitBitcoin(chainType);
 			InitLitecoin(chainType);
 			InitBCash(chainType);
-
+      InitPolis(chainType);
```
NicolasDorier: use tabs

cronos-polis:
Ok. Pull request is ready o  Test.Framework.
I already try to run it but got stuck on loading blocks from node.
Any help on how can I debug it?
Also, is there a need to change hash algorithm?

NicolasDorier:
@cronos-polis What hash algorithm you are talking about? If you are talking about the PoWHash, this is not needed. If you are talking about the way to caluclate the hash which each block use to reference each other, then it has to be SHA256d like bitcoin.

cronos-polis:
Polis is using X11 algorithm. I am pretty sure that is an issue. I already have this kind of proble.s when trying to port. 
Any way to implement it? I am stuck on the loading chain from node. Also when I read the information from the node I get the block height on 0. So the software is not recognizing the next block.

NicolasDorier:
Litecoin is using Scrypt for the PoW algorithm, but not for block linking. If POLIS is using X11 for block linking (sincerely, I doubt it though) then there is no way you can make it works now. 

We are doing some work on NBitcoin side to make it more flexible, but still not here yet.

cronos-polis:
Ok. Let's assume is not using. What else may I check. Or how can I debug it?

NicolasDorier:
I don't know where it stops. You can put breakpoint in the BitcoinDWaiter to see where it is stopping, or inside the ExplorerBehavior.

NicolasDorier:
@cronos-polis can you port the Network class in https://github.com/MetacoSA/NBitcoin/tree/master/NBitcoin.Altcoins ? I refactored NBitcoin yestersday to make it easier to integrate for alcoins.

There is some simple tests you can run which poke around if you did things right.

NicolasDorier:
You need to integrate regtest though.

cronos-polis:
Thanks.
I will do it.

cronos-polis:
Changed to NBitcoin.

PR #37, opened 2018-04-11
Fixed bug in Dogecoin initalization

bdangh:
Fixed misspelling which lead always initializing Regnet for Dogecoin

I still think doge synchronization not working on mainnet, as header format was changed after 371,337 block.

https://github.com/dogecoin/dogecoin/blob/master/doc/release-notes/RELEASE_NOTES_1_8.1.md


NicolasDorier:
There is also no doc anywhere on their block header format.

PR #41, opened 2018-04-13
Dash min rpc version and Doge timeout

bdangh:
Dash version 0.12.0.0 works fine so I decreased MinRPCVersion
Doge chain has 2,176,000+ blocks, so it takes more than 10 minutes to sync headers, temproarly increased timeout to one hour, until other solution founded.


NicolasDorier:
Can you rather add a new property to NBXplorerNetwork for the timeout with the default to 15min?

bdangh:
Sure, good idea, I will open new pull request

Issue #38, opened 2018-04-12
Is it possible to get back the ExPubKey from a bitcoin address?


bazooka70:
After creating a HD ExPubKey and registering the derivation scheme via `ExplorerClient.Track(userDerivationScheme)` is it possible to get the derivation scheme back (i.e. xpubkey)  from a given key path (which derived a bitcoin address ) from the NBXplorer database?

To Be more clear, suppose I have a bitcoin address e.g. "mnWoi6RjtvwHHRCHYmzzr14fZFf61drfA4" which was derived from path "0/0" (or "0/1" etc) can I somehow get back the "tpubD6NzVbkrYhZ4XtRwac3285m5k23....."? so I can tell to which HD wallet the address belongs?

I know it's a long shot but maybe the data base can return this information?

An alternative is querying and returning all the `userDerivationScheme` that were registered via `ExplorerClient.Track()`.

Thanks.

NicolasDorier:
The API does not expose this right now but I don't see any problem into adding a route for such feature.

bazooka70:
Great. I think this could be a very usul feature!

NicolasDorier:
no, you should submit a tested PR and then close :p

bazooka70:
Baaaaaaahhh! How on earth should I know where to start dinging with this issue? can you please give me a lead?

NicolasDorier:
need to update `MainController`, `NBXplorerClient` and then test it by adding a line somewhere in any Tests called `Track`.

bazooka70:
I think this is way over my head. I'm not sure which table holds that information. 
After reading the code a thousand times, my guess was `HighestPath` where it stores 

In the `MainControler` I added:

```csharp
                [HttpGet]
                [Route("cryptos/{cryptoCode}/derivations/getderivations")] 
		public IActionResult GetDerivations(
			string cryptoCode)
		{
			var network = GetNetwork(cryptoCode);
			var repo = RepositoryProvider.GetRepository(network);
			var tracked = repo.GetTracked();
			return Json(tracked);
		}

```

In the `Repository.cs` I added to the `Index` class
```csharp
                        public IEnumerable<DBreeze.DataTypes.Row<string, T>> SelectForward<T>(DBreeze.Transactions.Transaction tx) 
			{
				return tx.SelectForward<string, T>(TableName);
			}
```

In the `Repository` class I added:

```csharp
                public KeyPathInformation[] GetTracked()
		{			
			_Engine.Do(tx =>
[...]

PR #44, opened 2018-04-16
issue #38

bazooka70:
https://github.com/dgarage/NBXplorer/issues/38

Added `ExplorerClient.GetDerivations()`
You can query the available derivations via:
HTTP GET v1/cryptos/{cryptoCode}/derivations/getderivations

optional parameters:
```
DerivationStrategyBase strategy = null, 
DerivationFeature? feature = null,
KeyPath keyPath = null, 
Script scriptPubKey = null
```

Examples:

To query all derivations with KeyPaths of "0/0" use:

`var derivations = client.GetDerivations(null, null, new KeyPath("0/0"), null);`

To query for a bitcoin address use for e.g.:

`var derivations = client.GetDerivations(null, null, null, publicAddress.ScriptPubKey);`

Any combination of the optional parameters will use **AND** operator.


NicolasDorier:
that seems awesome, can you add a small test ?
No need of something fancy, you can hijack one of the existing tests if you want.

bazooka70:
OK, the tests are good. tests passed.

NicolasDorier:
very good! Will test / merge and modify a bit today. :)

bazooka70:
Thanks :) It was good I dived into a fragment of your code. your work is outstanding!

NicolasDorier:
Will do tomorrow actually, there is some small thing I want to modify after merging this.

You added lot's of query capabilities, sadly NBXplorer depends on NoSQL DB which is quite inflexible for queries. The way you are fetching the key is by enumerating all the table. This is very inefficient and would make NBXplorer crash if too many people use it.

The only query you can do is on the scriptPubKey. Because the scriptPubKey is a primary key.

bazooka70:
I know exactly what you mean, but I had no other choices if I wanted to make the query flexible/generic but to load the entire table first. 

Hell, I even wanted to make a `HashSet` for each optional parameter (which is easy to implement), but left the idea to later.

Its a Key-Value db, and problem is that you can't query on the value objects directly.
For each available derivation combination it produces 90 rows. 
So if we have many wallets to track, we load a LOT of rows into memory every time we query even a for a single derivation with a potential to crash. So yes, it's inefficient.

In short, I trust you know better and if you think you can optimize please do so. You can dump the whole idea if you think its not worth the trouble. The last thing I want is a crash because of this feature.

NicolasDorier:
What you wanted to use was `Repository.GetKeyInformations` I don't think you have to modify the Repository at all.

bazooka70:
So you suggest to query only by Scripts[]?

bazooka70:
Something like this?
```csharp
                [HttpGet]
		[Route("cryptos/{cryptoCode}/getkeyinformations")] 
		public IActionResult GetKeyInformations(string cryptoCode,
			[ModelBinder(BinderType = typeof(ScriptsModelBinder))]
			HashSet<Script> scripts
			)
		{
			var network = GetNetwork(cryptoCode);
			var repo = RepositoryProvider.GetRepository(network);
			var keyInformations = repo.GetKeyInformations(scripts.ToArray());
			var result = new List<KeyPathInformation>();
			foreach (var keyInfoByScripts in keyInformations)
			{
				foreach (var keyInfo in keyInfoByScripts.Value)
				{
					result.Add(keyInfo);
				}
			}			
			return Json(result);
		}
```

NicolasDorier:
Fix:

```
[Route("cryptos/{cryptoCode}/scripts/{script}")] 
```

The paramater should be one `Script script`. The ModelBinder should just read it as hex and deserialize it into a Script.

bazooka70:
Can the `repo.GetKeyInformations()` return more than a single element for one `Script`? 
I read this comment:

> // Because xpub are mutable (several xpub map to same script)
> // an attacker could generate lot's of xpub mapping to the same script
> // and this would blow up here. This we take only 5 results max.

And I'm not sure I follow how is that possible...

NicolasDorier:
Yes one can generate infinite `xpub` which create all same script. The reason is that not all info inside one xpub it relevant to derivation.

IMHO you should just returns either the array or the first result after having stripped the ExtPubKey from the KeyInformation so consumer's can't use it and can't shoot themselves in the foot by relying on it.

bazooka70:
I still don't get it. how can you have several KeyInformation for a single Script which was derived from a Keypath and suppose to be unique for the cryptocoin/network?
Can you give me a real example?

> after having stripped the ExtPubKey from the KeyInformation...

But this is exactly what I wanted to return in the first place.

Or you mean to say that there can be a situation where you have derivation of:
`tpubD6NzVbkrYhZ4XtRwac...-[legacy]`
`tpubD6NzVbkrYhZ4XtRwac...-[p2sh]`
etc..
which map to the same Script?

bazooka70:
This is what I came up with:

```csharp
                [HttpGet]
		[Route("cryptos/{cryptoCode}/scripts/{script}")] 
		public IActionResult GetKeyInformations(string cryptoCode,
			[ModelBinder(BinderType = typeof(ScriptModelBinder))] Script script)
		{
			var network = GetNetwork(cryptoCode);
			var repo = RepositoryProvider.GetRepository(network);
			var scripts = new Script[] { script };
			var keyInformations = repo.GetKeyInformations(scripts);			
			var result = new List<KeyPathInformation>();			
			foreach (var keyInfoByScripts in keyInformations)
			{
				foreach (var keyInfo in keyInfoByScripts.Value)
				{
					result.Add(keyInfo);
				}
			}
			if (result.Count == 0)							
				return NotFound();			
			return Json(result.ToArray());
		}
```

ScriptModelBinder:

```csharp
namespace NBXplorer.ModelBinders
{
	public class ScriptModelBinder : IModelBinder
	{
		#region IModelBinder Members

		public Task BindModelAsync(ModelBindingContext bindingContext)
		{
			if (!typeof(Script).GetTypeInfo().IsAssignableFrom(bindingContext.ModelType))
			{
				return Task.CompletedTask;
			}

			ValueProviderResult val = bindingContext.ValueProvider.GetValue(
				bindingContext.ModelName);
			if (val == null)
			{
				return Task.CompletedTask;
			}

			string key = val.FirstValue as string;
			if (key == null)
			{
				bindingContext.Model = null;
				return Task.CompletedTask;
			}

			var value = new Script(Encoders.Hex.DecodeData(key));

			bindingContext.Result = ModelBindingResult.Success(value);
			return Task.CompletedTask;
		}

		#endregion
	}	
}
```

NicolasDorier:
Seems good!

```
xpub661MyMwAqRbcFrVYKi4Gwhhx3Qhf8xMMjTN6RzdBFC7NsGto96G3dPy2fssQoaytHbj7qR7WBs5GFTDvPHuSLKgER36RrSdnpBX2KCsS9Qg
```

and 

```
xpub69mdgvyDG2wbyPkUPgyoZyEddrXYYQaFsFcuBiNWb4qBj6nRDfGjCm8EB2R6HTnpAzbR7Z2vhwVPcQCY81S2T9t6aLkHF7ECTeBvRECTyk4
```

Map to same derivations. Because the depth value is not implicated in the derivation but influence the base58 format.

bazooka70:
Ahhh? I need to check the above this is new to me...
In any case I think I will return an array so everyone but me can shoot themself in the leg :-p
I cracked the code and it works fine.
Do I open a new PR or delete the shit I did or what?

NicolasDorier:
as you want, you can make new PR or update this one by pushing your work to your repo/branch.

PR #51, opened 2018-04-21
Adds BlockId to the Transaction result output

replaysMike:
It's provided in a NewTransactionEvent but was missing here. Shouldn't be any reason to exclude it that I could see, and it's needed when querying Transactions for confirmations where no event was received.


NicolasDorier:
Thanks!

PR #61, opened 2018-06-21
Add Viacoin

romanornr:


NicolasDorier:
You probably did not ran the tests, because I forgot to update the test framework with your fix on the other repo.

NicolasDorier:
I just updated the test framework package and referenced it in NBXplorer, can you check the tests pass with Viacoin ?

NicolasDorier:
I tried to run the tests, it failed. It is probably a bug in your NBitcoin integration. You need to run the NBitcoin Altcoins tests with Viacoin to find out why.

PR #63, opened 2018-06-27
Reverse Feathercoin fix

ChekaZ:
- Cookie Auth now enabled.


NicolasDorier:
I will need to push new package of NBitcoin.Altcoins and update reference here.

Issue #73, opened 2018-07-25
Tracking addresses that already have balance?


bazooka70:
Hi @NicolasDorier , I am now facing a very hard issue. We stared working with Trezor wallets. the Trezor gives us an XPUB-key, and I register it with NBXplorer via `client.Track()`. All is well. 

The problem is with Trezor devices that already have balance **before** the `Track()` was called. 
No Balance and no transactions are returned.

Is there any way to re-sync the XPUB-key so that NBXplorer can give me the correct details for an address that had already a balance/transactions?

I know this is not the way NBXplorer suppose to work. but I really hate to switch to QBitNinja or other external indexers. (That is a real pain since the Trezor is an HD wallet and each wallet has multiple address).

Is there anything could be done?

NicolasDorier:
Yes, NBXplorer does not scan anything from the before it was first run.

You can ask for rescan with `--btcrescan=1` and `--btcstartheight=x` or something like this. Check with `--help`.

That said it can take a while.

bazooka70:
Yes, But even if that works I need to restart NBXplorer :/
Is this could be done (in theory) while the NBXplorer is running?

NicolasDorier:
If you use an external block explorer to fetch the UTXOs of newly tracked address, it would works fine while running.

bazooka70:
Can you please explain what you mean?

bazooka70:
I need to digest what you just said... 
I presume you expect a PR? :-p

NicolasDorier:
yes. You can hard code qbitninja for now if you want, I will cleanup later.

bazooka70:
I still need to get my shit together with QBitNinja (or other block explorer maybe?). 
The Trezor wallets are HD, and it allows segwit addresses. getting all the transactions seems to be very slow with QBitNinja. I need to run more tests. *sigh*

Another options (I think) might be to create a route that will only insert the data into the Repository for such situation, so the block explorer could be external and not hard-coded into NBXplorer... maybe.

NicolasDorier:
> Another options (I think) might be to create a route that will only insert the data into the Repository for such situation, so the block explorer could be external and not hard-coded into NBXplorer... maybe.

You are right this would be way better.

NicolasDorier:
Getting all transactions should not be that slow if you parallelize properly with Tasks. Also it is one time operation.

bazooka70:
Can please give me some leads as where to start? what table(s) should be updates? etc...

[...]

PR #80, opened 2018-08-06
Tracking addresses that already have balance #73

bazooka70:
Tracking addresses that already have balance.
PR in progress.

### DIFF
--- a/NBXplorer.Client/ExplorerClient.cs
+++ b/NBXplorer.Client/ExplorerClient.cs
@@ -210,6 +210,19 @@ namespace NBXplorer
 		{
 			return SendAsync<string>(HttpMethod.Post, keyPaths, "v1/cryptos/{0}/derivations/{1}/addresses/cancelreservation", new[] { CryptoCode, strategy.ToString() }, cancellation);
 		}
+		
+		public SaveTransactionsResult SaveTransactionsIds(DerivationStrategyBase strategy, uint256[] txIds, bool track = true, CancellationToken cancellation = default) //@@@
+		{
+			return SaveTransactionsIdsAsync(strategy, txIds, track, cancellation).GetAwaiter().GetResult();
+		}
+		public Task<SaveTransactionsResult> SaveTransactionsIdsAsync(DerivationStrategyBase strategy, uint256[] txIds, bool track = true, CancellationToken cancellation = default) // @@@
+		{			
+			Dictionary<string, string> parameters = new Dictionary<string, string>();
+			parameters.Add("txIds", String.Join(",", txIds.Select(t => t.ToString())));
+			parameters.Add("track", track.ToString());
+			var query = String.Join("&", parameters.Select(p => p.Key + "=" + p.Value).ToArray());			
+			return SendAsync<SaveTransactionsResult>(HttpMethod.Get, null, $"v1/cryptos/{CryptoCode}/derivations/{strategy}/transactions/savetxids?" + query, null, cancellation);		
+		}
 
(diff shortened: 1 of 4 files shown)


bazooka70:
Done

NicolasDorier:
Cool. Question though: Since you can get the txid from an external source, can this external source also provide you with the transactions ?

bazooka70:
Well yes I suppose it can with much more efforts. for example blockchain info api gives you only tx block height so you need a nother api to get block hash. if blockchain info is unavailable I switch to other apis and so on. and each one of them is totally different. so constructing  Transactions miight be a pain in the ass.

NicolasDorier:
Thanks a lot, I will likely change a bit this function to be able to pass the transactions, while still letting the capabilities of let them blank in case where the user has `txindex=1`.

bazooka70:
Will I still be able to pass only tx ids?

NicolasDorier:
Yes you will be able to pass only the tx ids.

bazooka70:
Thanks man. That sounds awesome!

NicolasDorier:
Sure I will make a PR so you can review.

bazooka70:
Never mind the PR.  I was just curious at how the method parameters will look like... I'm sure you know way better than me at how to implement this.

bazooka70:
OK, fair enough.

bazooka70:
Any progress?

NicolasDorier:
Sorry I did not had time, will do today

NicolasDorier:
So it turns out, that we can remove the need of `-txindex` if we pass the `blockid` of each `txid`. Would it be fine for you?

bazooka70:
Not sure I follow. Where do you see a `txindex` in my code?

NicolasDorier:
Thanks for your help,

https://github.com/dgarage/NBXplorer/commit/88b11589d3c7b06c41b861d338207a9191c3007b

I just pushed package and docker. What you did will still work. (API is different though, but the inputs are the same if you have `-txindex`)

I extended a bit, and I batch RPC requests so it is faster.

I also test it.

bazooka70:
Thank you! I have much to learn from you.
BTW, if the tx is not confirmed yet, you must use the `--txindex` anyway. or am I wrong?

NicolasDorier:
I did not tried rescan on unconfirmed transaction. It might crash. That said this should not happen, because this is in general for old transaction.

bazooka70:
@NicolasDorier , Just to be 100% clear: calling this methods multiple times (`Track` & `Rescan`), will make no harm? no duplicate data will be inserted to the DB?

PS: do you by chance know any DBreeze desktop explorer?

NicolasDorier:
@bazooka70 no DBreeze explorer, but no worry, it should not insert twice. You can try.

bazooka70:
@NicolasDorier, yes I have tested and it *seems* there are no duplicates, and everything really works like a charm! :100:  
Does the DBreeze ignore inserting a duplicate key(s)? or there is an exception which you internally ignore/suppress?

bazooka70:
OK thanks! moving to production... :)

bazooka70:
BTW, the old versions had `chain-stripped.dat` but now it keeps `chain-slim.dat`.

What will happen if if update the NBXplorer server on a machine which already was running old version of NBXplorer and has `chain-stripped.dat`?

NicolasDorier:
don't worry about this data, it is just an optimization for starting nbxplorer faster. Not important data.

bazooka70:
OK, NBXplorer deleted it for me :)
Running on production. so far so good! :1st_place_medal:

PR #83, opened 2018-08-13
Update README.md

chriswhellams:
Fixed typos, added a section on using Postman with Basic Auth to test the API , with example.


NicolasDorier:
Excellent, though better way to run is to use the `run.sh` or `run.ps1` script.

PR #86, opened 2018-08-20
Issue#85 - Message Broker support

chriswhellams:
- Added support for Azure Service Bus message broker as IHostedService
- Added 2 Unit tests for Azure Service Bus and static TestConfig.cs to supply config settings
- Extended config settings to support Azure Service Bus config
- Removed some unused using statements
- Added Azure Service Broker extension for IServiceCollection / DI / IoC support
- Updated readme.md with information on Azure Service Bus Message Broker and how to configure

- All unit tests passing when active Service Bus connection string supplied. 
- Azure service bus tests fail with no connection string message when connection string unavailable

### DIFF
--- a/NBXplorer.Tests/TestConfig.cs
+++ b/NBXplorer.Tests/TestConfig.cs
@@ -1,0 +1,34 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace NBXplorer.Tests
+{
+	public static class AzureServiceBusTestConfig
+	{
+		public static string ConnectionString
+		{
+			get
+			{
+				//Put your service bus connection string here - requires READ / WRITE permissions
+				return "";
+			}
+		}
+
+		public static string NewBlockQueue
+		{
+			get
+			{
+				return "newblock";
+			}
+		}
+
+		public static string NewTransactionQueue
+		{
+			get
+			{
+				return "newtransaction";
+			}
+		}
+	}
+}
--- a/NBXplorer/Extensions.cs
+++ b/NBXplorer/Extensions.cs
@@ -157,16 +157,6 @@ namespace NBXplorer
 			});
 		}
 
-		public static IServiceCollection AddMessageBrokers(this IServiceCollection services,IConfiguration config )
-		{
-			//There must be a better way to do this ! IoC container should be able to inject config.
-
-			if (!string.IsNullOrWhiteSpace(config["asbcnstr"]))
-				services.AddSingleton<IHostedService, AzureServiceBus>();
-
-			return services;
-		}
-
 
 		public static IServiceCollection AddNBXplorer(this IServiceCollection services)
 		{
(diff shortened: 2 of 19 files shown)


NicolasDorier:
Nice nice! 
I apologize, yestersday I did some changes and broke your PR, can you rebase/solve conflicts?

NBXplorer.Tests/TestConfig.cs
```diff
@@ -0,0 +1,34 @@
...
+		public static string ConnectionString
+		{
+			get
+			{
+				//Put your service bus connection string here - requires READ / WRITE permissions
+				return "";
+			}
+		}
+
+		public static string NewBlockQueue
```
NicolasDorier: Use public const string NewBlockQueue = "newblock";

NicolasDorier: Actually you can remove that right? this seems to be the default

NicolasDorier:
I would be curious if you can use Topic with it, if you can't I don't think the code change should be that big.

NBXplorer/Extensions.cs
```diff
@@ -156,6 +157,17 @@ public static AuthenticationBuilder AddNBXplorerAuthentication(this Authenticati
 			});
 		}
 
+		public static IServiceCollection AddMessageBrokers(this IServiceCollection services,IConfiguration config )
+		{
+			//There must be a better way to do this ! IoC container should be able to inject config.
+
+			if (!string.IsNullOrWhiteSpace(config["asbcnstr"]))
+				services.AddSingleton<IHostedService, AzureServiceBus>();
```
NicolasDorier: Please add this inside AddNBXplorer, Always add this hosted service. Just make sure that it does nothing when it starts if no connection strings have been set. (As you can resolve the configuration object in the ctor of the class)

NBXplorer/MessageBrokers/AzureServiceBus.cs
```diff
@@ -0,0 +1,238 @@
...
+		public async Task StartAsync(CancellationToken cancellationToken)
+		{
+			if (_Disposed)
+				throw new ObjectDisposedException(nameof(AzureServiceBus));
+
+
+			string listenAllDerivationSchemes	= null;
+			var	listenedBlocks					= new ConcurrentDictionary<string, string>();
+			var listenedDerivations				= new ConcurrentDictionary<(Network, DerivationStrategyBase), DerivationStrategyBase>();
+
```
NicolasDorier: I think you don't need all of that, the client can filter by himself.

chriswhellams:
Thanks for code review and feedback. I can add topics, it's not that hard. I'll make all the changes and rebase.

NBXplorer.Tests/TestConfig.cs
chriswhellams: I think these should stay in TestConfig as the Test will independently connect to the queues and does not have access to the defaults set in NBXplorer config. If  these values are hardcoded in the test then you are forced to name your queues in your test environment with the hardcoded values.

chriswhellams:
@NicolasDorier I sent a new PR as it was easier to clone a new copy of your master and make my changes on that.

Issue #85, opened 2018-08-14
Feeding NBXplorer notifications to RabbitMQ


sigHash:
Hi! I am building a worker (.net core console) that uses NBXplorer.Client to listen for NBXplorer notifications, process each event and send them over to RabbitMQ.  

Should I be worried about buffering those events while I am processing them, or the client will take care of that for me? The assumption is that I'm subscribing to thousands of newTransaction and NewBlock events. 

Thanks!

NicolasDorier:
Are you sure you will process that much? Because it sends events only for transactions which match a xpub.

I don't know how buffering work as far as websocket are concerned. You would have to try and see what happen if events keep coming but the consumer does not process them.

NicolasDorier:
Problably a better solution for you might be to change NBXplorer to publish notifications on RabbitMQ directly.

NicolasDorier:
Well I am thinking merging it on this repo. 

Create a HostedService which listen the proper events on the Event Aggregator and forward it to RabbitMQ. (It is quite easy, check what I did for websocket in the MainController)

It is probably easier than websocket.

sigHash:
Got it! Just checked the code on [Route("cryptos/{cryptoCode}/connect")].

Now I'm thinking if we have multiple instances of NBXplorer, we'll end up with many duplicate messages in RabbitMQ.

chriswhellams:
Great idea. It would be really nice if this could be done either with AMQP 1.0 or as a generic interface that would allow others to extend. Reasons:

1. AMQP 1.0 is also supported by e.g. Azure Service Bus and is a standard. Negative - RabbitMQ is based on AMQP 0.9.1 and requires a plugin for AMQP 1.0 . https://www.rabbitmq.com/protocols.html

2. (And / or )abstract implementation with an interface , allowing others to add support for other brokers such as ZeroMq, Azure service bus, MQTT etc . 

@sigHash  you might know this already - but bitcoind already has ZeroMQ support  https://github.com/bitcoin/bitcoin/blob/master/doc/zmq.md but lacks the nice filtering in NBXplorer.

sigHash:
@chriswhellams I agree. WS doesn't work for us either. De-duplication is a real issue if new messages trigger expensive updates, but you are right; It's a separate layer and I am not too worried about it because you can flush em out.

I'll create the HostedService and cout events for now. You can then use your cloud provider SDKs to send them wherever you want.

[...]

PR #87, opened 2018-08-22
Added Azure Service Bus support

chriswhellams:
- Added support for Azure Service Bus message broker as IHostedService. Supports both Queues and Topics.

- Uses same serialization as WebSocket

- Moved ToTransactionResult from MainController to Utils to aid re-use

- Added NewEventBase abstract base class with helper methods for Json conversion, updated NewEvent classes to derive from base.

- Added 2 Unit tests for Azure Service Bus and static TestConfig.cs to supply config settings

- Extended config settings to support Azure Service Bus config

- Removed some unused using statements

- Updated readme.md with information on Azure Service Bus Message Broker and how to configure

- All unit tests passing when active Service Bus connection string supplied.

- Azure service bus tests fail with no connection string message when connection string unavailable

- Rebased from latest Main branch, no merge conflicts

### DIFF
--- a/NBXplorer.Client/Models/NewBlockEvent.cs
+++ b/NBXplorer.Client/Models/NewBlockEvent.cs
@@ -1,12 +1,9 @@
 ﻿using NBitcoin;
-using System;
-using System.Collections.Generic;
-using System.Text;
 
 namespace NBXplorer.Models
 {
-    public class NewBlockEvent
-    {
+	public class NewBlockEvent : NewEventBase
+	{
 		public int Height
 		{
 			get; set;
@@ -19,11 +16,6 @@ namespace NBXplorer.Models
 		public uint256 PreviousBlockHash
 		{
 			get; set;
-		}
-		public string CryptoCode
-		{
-			get;
-			set;
 		}
 	}
 }
--- a/NBXplorer.Client/Models/NewBlockEventRequest.cs
+++ b/NBXplorer.Client/Models/NewBlockEventRequest.cs
@@ -1,20 +1,7 @@
-﻿using NBitcoin;
-using System;
-using System.Collections.Generic;
-using System.Text;
-
-namespace NBXplorer.Models
+﻿namespace NBXplorer.Models
 {
-    public class NewBlockEventRequest
-    {
-		public NewBlockEventRequest()
-		{
+	public class NewBlockEventRequest : NewEventBase
+	{
 
-		}
-
-		public String CryptoCode
-		{
-			get; set;
-		}
 	}
 }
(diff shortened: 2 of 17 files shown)


chriswhellams:
@NicolasDorier All requested changes - except hardcoding queue / topic names are implemented. No merge conflicts .

NicolasDorier:
You can dramatically simplify this code, Topic and Queue have a common interface `ISenderClient`.
Which mean that you don't have to have `_queueTran` and `_queueTopic` instead you can have one `_queueSender`.

NicolasDorier:
Do you want I cleanup or you want to try to do it?

vineeth7858:
No don't cleanup

chriswhellams:
@NicolasDorier  ooh, good find ! Msft have made the client libs for Azure Service Bus somewhat confusing as they have different versions for framework and core and docs are outdated or specific to one of the client implementations.  Let me take a look at this now and I can see how it works and I'll make the change and test everything again.

chriswhellams:
@NicolasDorier I looked at ISenderClient, I am not really seeing how this simplifies the code as it appears to be just an abstraction interface that derives from ITopicClient, IQueueClient and IMessageSender and provides the "lowest common denominator" from those interfaces - which is pretty much SendAsync. It still looks like you need to use concrete classes such as QueueClient and TopicClient . Queues and Topics are not the same thing so I don't think you can use a combined _queueSender. I cannot find any worked examples from Google of this being used anywhere - which is often a sign nobody is using it ! I am somewhat cautious / nervous of using Msft features that lack examples / documentation due to the fact Msft seem not to test much these days and let the community do that instead - the Msft PM responsible for service bus also seems to be far more interested in Event Hubs these days - which is also a warning sign for me of something un-loved.

Ok - you could make something fancy and inject in the concrete classes and use a single method to handle things but that will add complexity.

https://docs.microsoft.com/en-us/dotnet/api/microsoft.azure.servicebus.core.isenderclient?view=azure-dotnet

https://azure.microsoft.com/en-us/blog/author/juconway/

If you want to switch to ISenderClient it's your call, I'm not going to do that unless there is an obvious benefit or I have totally missed something. Have I totally missed something ? ;-)

NicolasDorier:
Ok, I'll take a look.

NicolasDorier:
My apologies, I completely forgot about it, working on it now.

NicolasDorier:
I took the freedom to cleanup your code on ac905261cbbdd6bc37822fa513a8f67e0255ddaa .

Basically, you were copying 4 times the same code for Queue/Topic, Transaction/Block, which was not good. The tests were not using assert, making it hard to debug.

I also prepared the way for RabbitMQ support to which @sigHash seems interested it https://github.com/dgarage/NBXplorer/issues/85 by abstracting azure specific stuff into a `IBrokerClient` interface fairly easy to implement.

chriswhellams:
Many thanks for taking the PR and for improving.

PR #89, opened 2018-08-30
WIP:Feature/mass transit

Kukks:
This expands the new ASB support to also have 2 MassTransit brokers support:
ASB via MassTransit
RabbitMQ via MassTransit

Missing from this:

- [x] Integrate into `BrokerHostedService`
- [x] Fetch config from the `ExplorerConfiguration`
- [ ] Unit Test MassTransit brokers


Kukks:
yup, forgot to stage those :)

chriswhellams:
Nice, we used MT in a previous job. It's a good framework.

NicolasDorier:
love it

NicolasDorier:
can you provide a docker-compose which fireup rabbitmq so we can test ?

NicolasDorier:
`blockbroker.0.broadcasttype=send` make no sense for asb

Kukks:
Added it in the regtest docker @NicolasDorier

NicolasDorier:
I don't understand the `broadcasttype` concept, what is it?

Kukks:
Publish is when the message is sent to all subscribers and send is when it is sent to a queue and processed by one subscriber. 
Maybe I got the asb terminology mixed up a bit though? 
https://www.maldworth.com/2015/10/27/masstransit-send-vs-publish/

NicolasDorier:
it works with azure queue as well? But you need to set to send right?

Kukks:
Yeah, so send is the equivalent of a queue and publish is for the topic

NicolasDorier:
It seems good to me, can you update the doc?

Kukks:
Updated readme and the default config creator. I had to remove the asb arguments as it is a bit too complex for that now. Just needs tests now. :)

NicolasDorier:
still need tests

Kukks:
Yes, started adding some a few days ago, will finish up soon

Kukks:
@NicolasDorier  I added in some tests but whenever I try to test the `ServerTester.Create()` seems to never finish executing?

NicolasDorier:
@Kukks it means you are not properly disposing something

Kukks:
It seems to be something with the `NodeBuilder` as it never gets past that?

Kukks:
@NicolasDorier  Do you have any docs somewhere on how to run the tests? I haven't been able to get past this issue

NicolasDorier:
It is because it download bitcoin core so the first run might take a while.

Kukks:
I need to revisit this  on a separate pull request, but maybe not at all since nobody seems to actually need it.

PR #102, opened 2018-09-17
Hash Azure Serive Bus messageId and Validate

chriswhellams:
Azure Service Bus message id's cannot be null and must be 128 character string or less (based on C# client library validator).

The message Id is made by concatenating up to 3 hashes - this can result in > 128 chars. This PR creates a hash of the messageId to create a shorter messageId and also adds a validator to check the message id is valid before broadcasting.


NicolasDorier:
Why closed?

chriswhellams:
I must have closed accidentally - I committed some test config settings and tried to clean up

NicolasDorier:
yes

PR #103, opened 2018-09-18
Issue101 - Fix Azure service Bus MessageId length

chriswhellams:
MessageId in azure service bus client cannot be null or empty and cannot exceed 128 characters. This change hashes the unique MessageId to shorten it and then performs validation


NicolasDorier:
Why the test did not catch that?

PR #137, opened 2018-11-22
CircleCI and building of Docker images

rockstardev:
Take into account that recent changes on master now cause 2 tests to fail:

  - CanSendAzureServiceBusNewTransactionEventMessage
  - CanSendAzureServiceBusNewBlockEventMessage

Otherwise everything is working as for BTCPayServer. Here is test docker repo as defined in $DOCKERHUB_REPO variable:

https://hub.docker.com/r/rockstardev/nbxplorer/tags/


rockstardev:
Could be those two are failing because of missing creds... I see you have appveyor building this, meaning you likely need to also port NuGet stuff.

But you can use these commits as starting point @NicolasDorier

PR #35, opened 2018-04-06
Dogecoin integration foundation

lepipele:
Does need review and additional work

PR #175, opened 2019-06-27
Update Regex

junderw:


NicolasDorier:
why? This regex is just for fetching the xpubs, I am parsing the xpub properly via NBitcoin once I get the strings.

junderw:
I have tested:

4 byte version ("xpub") + rest of bytes are 0x00.

4 byte version ("xpub") + rest of bytes are 0xff.

All xpubs are 111 characters exactly, sub 4 for xpub, so xpubs will always start with xpub with 107 base58 chars after it.

junderw:
> why?

Because it is correct.

For sure, now it causes no problems. But loose Regex can cause bugs down the line / open you up to some DDoS vectors with loose Regex down the line.

If matching in Regex, you want to match as exact as possible if you can.

15-of-15 is the largest possible multisig, so allowing 16 will cause an error or lost funds in some apps.

Obviously NBitcoin is smart, so it would never do that. I know.

But I am just making a fix I saw. It is not too important. But I developed a translating app in-house for years in my previous employer and I am trained like a dog to be very serious about Regex :-P

NicolasDorier:
It is not always xpub, each shitcoin have their own prefix, also we need tpub.

I prefer not merging this as the actual parsing is still done and would catch anything not properly formated.

junderw:
hmmm...

In which case it would be better to parse at a lower layer that understands shitcoins,

rather than making a if-else here, make the if-else in the NBitcoin logic as some sort ok HDMultisigKey class or something.

what if a shitcoin has xpub with a hyphen in it or has base64 symbols etc?

NicolasDorier:
> In which case it would be better to parse at a lower layer that understands shitcoins,

This is what it is doing.
The regex is only here so I can capture the xpubs into groups, then I parse each group via a lower shitcoin layer handled by NBitcoin.

NBitcoin already know how to parse any XPUB format internally. (Something like new `new BitcoinExtPubKey(str, network)` where network is the shitcoin/alternative network)

NicolasDorier:
You can see it

```csharp
var pubKeys = match.Groups
									.OfType<Group>()
									.Skip(2)
									.SelectMany(g => g.Captures.OfType<Capture>())
									.Select(g => new BitcoinExtPubKey(g.Value.Substring(1), Network))
									.ToArray();
```

junderw:
what if shitcoin xpub header is:

xshit:-%#35&

and because of this, NBitcoin change the assumption that all xpubs are base58?

NBitcoin should parse "2-of-xpub..." just as a string.

This regex here is under an assumption that all xpubs of all coins will always be alphanumeric.

junderw:
so as current, the "m-of-xpub" format is in NBXplorer layer, but I think it should be in NBitcoin layer, so shitcoins can also modify it.

junderw:
If it is NBXplorer layer, usually you want parsing and encoding logic separated so you can test the encode-decode separately.

junderw:
anywho, I am sure someday this Regex will be the cause of some bug when you refactor. :-P

this PR can be closed.

NicolasDorier:
> what if shitcoin xpub header is:

This shitcoin would not be supported, or I will change the regex at this point if needed.

NBitcoin does not know the concept of derivation scheme, nor should it because it is not a standard bip, this is a specific thing in NBXplorer.

PR #176, opened 2019-07-29
Fix regtest yml file

junderw:
I was having trouble getting regtest to work.

This fixed it...

Perhaps there is a better way of fixing this.


NicolasDorier:
@junderw yes, there is better way. Will do.

NicolasDorier:
Try https://github.com/dgarage/NBXplorer/commit/567cd245be0e8768ce2113c415fe528b85048ed9

junderw:
ok that works fine

PR #198, opened 2019-10-07
Add Liquid Support

Kukks:
This PR introduces support for the Liquid network and a few other minor refactorings.

* New network Liquid
* Makes the Serializer require the NBXplorer Network instead of NBitcoin network . This makes more sense within NBX's scope. It was needed so that the DerivationSchemeTrackedSource parser was able to parse properly based on network. 
* Make NBX use the derivation strategy factory instance inside the NBX network instead of creating a new instance eveywhere, all the time.
* Supports liquid network's confidential addresses using `-[blindingkey=hex]`. This is an extension only on elements specific code and does not pollute the rest of the codebase. If a matched transaction needs unblinding, we import the blinding key to elementsd and unblind via rpc. If the matched outputs are still unblinded, we ignore the tx.
* Bump NBitcoin to latest major version and fixes breaking changes.
* Add some generic client methods to ExplorerClient as there are derived types for Liquid specific code. 

I'm pretty happy with how this code turned out now, it keeps all of the non bitcoin code as separate and out of the main codebase as possible. The only liquid specific code I did not get rid of was this:
`var repo = net.NBitcoinNetwork.NetworkSet == Liquid.Instance ? new LiquidRepository(_Engine, net, keyPathTemplates, settings.RPC) : new Repository(_Engine, net, keyPathTemplates, settings.RPC);`
 and that was to avoid introducing some complex abstractions. ( I can always introduce a repository factory but I think it is over the top) 

There is a breaking change because of the signature change of `TrackedSource.TryParse` and `Serializer.ctor` of the parameter from `Network` to `NBXplorerNetwork` but it is worth it for the added flexibility.


NBXplorer/Repository.cs
```diff
@@ -76,7 +68,15 @@ public RepositoryProvider(NBXplorerNetworkProvider networks, KeyPathTemplates ke
 				var settings = GetChainSetting(net);
 				if (settings != null)
 				{
-					var repo = new Repository(_Engine, net, keyPathTemplates);
+					Repository repo;
```
Kukks: this section sucks but I cant move this logic anywhere else so far

NBXplorer.Client/DerivationStrategy/DerivationStrategy.cs
```diff
@@ -56,13 +58,13 @@ public DerivationStrategyFactory(Network network)
...
-		public DerivationStrategyBase Parse(string str)
+		public virtual DerivationStrategyBase Parse(string str)
 		{
-			var strategy = ParseCore(str);
+			var strategy = ParseCore(str, new Dictionary<string, object>());
 			return strategy;
 		}
 
-		private DerivationStrategyBase ParseCore(string str)
+		protected DerivationStrategyBase ParseCore(string str, Dictionary<string, object> additionalOptions)
```
NicolasDorier: Make `additionalOptions` default to null.

NicolasDorier:
Remove everything concerning the addition of Blinding Key. There is nothing in liquid that require changing the format of derivation schemes. The blinding key must be derived from the xpub, not passed separately.

NicolasDorier:
Remove the `AdditionalOptions` and the blinding key tag. As I said we will derive private key from the xpub, not doing so will give too much trouble.

NBXplorer.Tests/UnitTest1.cs
```diff
@@ -130,7 +135,7 @@ public void CanSerializeKeyPathFast()
 			using (var tester = RepositoryTester.Create(true))
 			{
 				var dummy = new DirectDerivationStrategy(new ExtKey().Neuter().GetWif(Network.RegTest)) { Segwit = false };
-				var seria = new Serializer(Network.RegTest);
+				var seria = new Serializer(GetNetwork(Network.RegTest));
```
NicolasDorier: why not `tester.Network` ?

NBXplorer/Startup.cs
```diff
@@ -54,7 +55,7 @@ public void ConfigureServices(IServiceCollection services)
 			builder.AddNewtonsoftJson(options =>
 			{
 				options.SerializerSettings.ContractResolver = new CamelCasePropertyNamesContractResolver();
-				new Serializer(null).ConfigureSerializer(options.SerializerSettings);
+				new Serializer(new NBXplorerNetwork(NBitcoin.Bitcoin.Instance, NetworkType.Mainnet)).ConfigureSerializer(options.SerializerSettings);
```
NicolasDorier: no, you can't do this.

Kukks: I added this after bumping NBitcoin. Apparently the serializers from NBitcoin now require that the network is not null?

NBXplorer.Tests/UnitTest1.cs
Kukks: fixed

Kukks:
> 
> 
> Remove the `AdditionalOptions` and the blinding key tag. As I said we will derive private key from the xpub, not doing so will give too much trouble.

Why? I've tested the current way and everything is working( also in BTCPay via https://github.com/btcpayserver/btcpayserver/pull/1118). 

I prefer having the default being unblinded as it's way easier for wallet support.

Plus the `-[blindingkey=hex]` logic works quite well, does not pollute the btc code and is easy to import into a wallet to view all confidential transactions with just one key that is provided explicitly instead of needing to export data from NBX with more code.

NBXplorer/Startup.cs
NicolasDorier: Checking why, but this is a bug of NBXplorer not of NBitcoin. The NBitcoin 5.0 became less forgiving into serializing stuff without knowing the Network. And for good reason, as you, as it catched the fact that in liquid, you can't possibly serialize properly the TxOut properly.

NicolasDorier: I pushed a fix on master, rebase your PR.

Kukks: done

NicolasDorier: so can you remove this change?

NicolasDorier:
I fixed something else in NBitcoin, you can rebase.

NBXplorer/Startup.cs
Kukks: ah I thought I did, will do now

Kukks:
> 
> 
> I fixed something else in NBitcoin, you can rebase.

done

NBXplorer/Startup.cs
Kukks: done

nharmon9:
Testing and every time I try to run with LBTC i get this:
```
info: Configuration:  LBTC: Testing RPC connection to http://localhost:18884/ 
info: Configuration:  LBTC: RPC connection successful
info: Configuration:  LBTC: Full node version detected: 189900
fail: Configuration:  LBTC: Failed to connect to RPC
```

Also if I can help in any way let me know

NBXplorer.Client/JsonConverters/CachedSerializer.cs
```diff
@@ -52,12 +52,13 @@ internal bool CanConvert(Type objectType)
...
 		}
-		public CachedSerializer(Network network)
+
+		public CachedSerializer(NBXplorerNetwork network)
 		{
 			if (network == null)
 				throw new ArgumentNullException(nameof(network));
-			cachedConverter.Add(new CachedConverter(new BitcoinStringJsonConverter(network)));
-			cachedConverter.Add(new CachedConverter(new DerivationStrategyJsonConverter(new DerivationStrategy.DerivationStrategyFactory(network))));
+			cachedConverter.Add(new CachedConverter(new BitcoinStringJsonConverter(network?.NBitcoinNetwork)));
```
NicolasDorier: NBitcoin does not support null network here, so should be `network.NBitcoinNetwork`

NBXplorer.Client/JsonConverters/CachedSerializer.cs
```diff
@@ -52,12 +52,13 @@ internal bool CanConvert(Type objectType)
...
+			cachedConverter.Add(new CachedConverter(new DerivationStrategyJsonConverter(network?.DerivationStrategyFactory)));
```
NicolasDorier: same

NBXplorer/TrackedTransaction.cs
```diff
@@ -101,6 +101,8 @@ internal void KnownKeyPathMappingUpdated()
 			for (int i = 0; i < Transaction.Outputs.Count; i++)
 			{
 				var output = Transaction.Outputs[i];
+				if(output.Value == null )
+					continue;
```
NicolasDorier: I think you can remove this.

NicolasDorier:
You can rebase, should be simple the conflicts here are changes I picked from your fork.

NicolasDorier:
How can I run my test suite on liquid?

Kukks:
There should be a docker compose service and a launch profile for lbtc

NicolasDorier:
@Kukks but why?

NBXplorer tests can just use the TestFramework.

NBXplorer.Client/NBXplorerNetworkProvider.Liquid.cs
```diff
@@ -0,0 +1,209 @@
...
+			}
+
+			public override ExplorerClient CreateExplorerClient(Uri uri)
+			{
+				return new LiquidExplorerClient(this, uri);
+			}
+			
+			public override BitcoinAddress CreateAddress(DerivationStrategyBase derivationStrategy, KeyPath keyPath, Script scriptPubKey)
+			{
+				if (derivationStrategy is IHasDerivationStrategyOptions strategyOptions &&
```
NicolasDorier: just do not support unblinded. This make this PR smaller.

NBXplorer.Client/NBXplorerNetworkProvider.Liquid.cs
```diff
@@ -0,0 +1,209 @@
...
+				{
+					return base.CreateAddress(derivationStrategy, keyPath, scriptPubKey);
+				}
+				var blindingPubKey = blindingKey.PubKey;
+				return new BitcoinBlindedAddress(blindingPubKey, base.CreateAddress(derivationStrategy, keyPath, scriptPubKey));
+			}
+
+			public static Key GenerateBlindingKey(DerivationStrategyBase derivationStrategy, KeyPath keyPath)
+			{
+				if (derivationStrategy == null)
```
NicolasDorier: why derivationStrategy would be null?

Kukks: There is one place at least: 
```
public async Task<MultiValueDictionary<Script, KeyPathInformation>> GetKeyInformations(Script[] scripts)
		{
			MultiValueDictionary<Script, KeyPathInformation> result = new MultiValueDictionary<Script, KeyPathInformation>();
			if (scripts.Length == 0)
				return result;
			foreach (var batch in scripts.Batch(BatchSize))
			{
				await _TxContext.DoAsync(tx =>
				{
					tx.ValuesLazyLoadingIsOn = false;
					foreach (var script in batch)
					{
						var table = GetScriptsIndex(tx, script);
						var keyInfos = table.SelectForwardSkip(0)
											.Select(r => GetKeyPathInformation(r.Value, null))
											// Because xpub are mutable (several xpub map to same script)
											// an attacker could generate lot's of xpub mapping to the same script
											// and this would blow up here. This we take only 5 results max.
											.Take(5)
											.ToArray();
						result.AddRange(script, keyInfos);
					}
				});
			}
			return result;
		}


		protected virtual KeyPathInformation GetKeyPathInformation(byte[] value, DerivationStrategyBase derivationStrategyBase)
		{
			var result = ToObject<KeyPathInformation>(value);
			if (result.Address == null)
			{
				result.Address = Network.CreateAddress(derivationStrategyBase, result.KeyPath, result.ScriptPubKey);
			}

			return result;
		}
```

NBXplorer.Client/NBXplorerNetworkProvider.Liquid.cs
Kukks: done, will do separate PR after

NBXplorer.Client/NBXplorerNetworkProvider.Liquid.cs
NicolasDorier: Before there was `AddAddress` with my change `AddAddress` will only apply to old KeyInfo.
New KeyInfo have their address saved in DB.

So this should never happen.

NBXplorer.Client/DerivationStrategy/Derivation.cs
```diff
@@ -19,5 +19,10 @@ public Script Redeem
 		{
 			get; set;
 		}
+
+		public virtual BitcoinAddress GetAddress(Network network)
```
NicolasDorier: This not needed anymore

NBXplorer.Client/DerivationStrategy/DerivationStrategy.cs
```diff
@@ -41,13 +41,13 @@ public DerivationStrategyFactory(Network network)
 
 		readonly Regex MultiSigRegex = new Regex("^([0-9]{1,2})-of(-[A-Za-z0-9]+)+$");
```
NicolasDorier: changes not needed to this file

NBXplorer.Client/ExplorerClient.cs
```diff
@@ -125,9 +125,14 @@ public Task<UTXOChanges> GetUTXOsAsync(DerivationStrategyBase extKey, Cancellati
 			return GetUTXOsAsync(TrackedSource.Create(extKey), cancellation);
 		}
```
NicolasDorier: I would prefer to not do this changes. This is over abstraction IMHO.

NBXplorer/Extensions.cs
```diff
@@ -78,15 +79,7 @@ public static T As<T>(this IActionResult actionResult)
 			}
 			return null;
 		}
-
-		internal static KeyPathInformation AddAddress(this KeyPathInformation keyPathInformation, Network network)
```
NicolasDorier: I don't think you need to refactor this part.

NBXplorer/Repository.cs
```diff
@@ -321,13 +315,14 @@ public NBXplorerNetwork Network
 		}
 
 		DBriizeTransactionContext _TxContext;
-		internal Repository(DBriizeEngine engine, NBXplorerNetwork network, KeyPathTemplates keyPathTemplates)
+		internal Repository(DBriizeEngine engine, NBXplorerNetwork network, KeyPathTemplates keyPathTemplates, RPCClient rpcClient)
```
NicolasDorier: I was considering it, but this is not needed anymore. I am importing in the GetUnused inside the controller. This create less changes.

Kukks: This is needed to unblind a transaction not just to import though

NicolasDorier: no, you can just unblind in the Behavior and only save the unblinded transaction  in the DB

NBXplorer.Client/NBXplorerNetworkProvider.Liquid.cs
Kukks: Ok,

NBXplorer/Extensions.cs
Kukks: Ok

NBXplorer/Repository.cs
Kukks: wait nvm, cleaning it

NBXplorer.Client/ExplorerClient.cs
Kukks: `GetTransactionAsync` is the only one needed now

NBXplorer/TrackedTransaction.cs
Kukks: This is in case the matched output was not blinded. Basically, if you have another wallet generating keys on the same paths with the same xpub, and uses a different blinding key, this will save invalid(impartial) data

NBXplorer.Client/NBXplorerNetworkProvider.Liquid.cs
```diff
@@ -0,0 +1,77 @@
...
+		}
+
+		public class  LiquidTransactionResult : TransactionResult
+		{
+			private ElementsTransaction _transaction;
+			public override Transaction Transaction
+			{
+				get => _transaction;
+				set => _transaction = value as ElementsTransaction;
+			}
```
NicolasDorier: this is completely useless.

NBXplorer.Client/ExplorerClient.cs
NicolasDorier: This is also completely useless as NBitcoin already serialize/deserialize the correct type.

NBXplorer.Client/JsonConverters/TrackedSourceJsonConverter.cs
```diff
@@ -11,12 +11,12 @@ namespace NBXplorer.JsonConverters
 {
 	public class TrackedSourceJsonConverter : JsonConverter
 	{
-		public TrackedSourceJsonConverter(Network network)
+		public TrackedSourceJsonConverter(NBXplorerNetwork network)
```
NicolasDorier: is it needed?

NBXplorer/TrackedTransaction.cs
NicolasDorier: it does not matter, we should still index that. That people can't read the value is their problem.

NicolasDorier:
Can you rebase, I integrated some of your changes in separate commit as it was a bug fix unrelated to this PR https://github.com/dgarage/NBXplorer/commit/51cd86d860639f280e7da478c5d370545ee0f11c

NBXplorer.Client/NBXplorerNetworkProvider.Liquid.cs
Kukks: No, without this, the Tx in BTCPay would only be `Transaction` without the required data

NicolasDorier: Nop

Kukks: you were right :)

NBXplorer/TrackedTransaction.cs
Kukks: done

NBXplorer.Client/Models/KeyPathInformation.cs
```diff
@@ -24,8 +24,18 @@ public KeyPathInformation(DerivationFeature feature, KeyPath keyPath, Derivation
 			KeyPath = keyPath;
 			Address = network.CreateAddress(derivationStrategy, keyPath, ScriptPubKey);
 		}
+		
+		public KeyPathInformation(Derivation derivation, DerivationSchemeTrackedSource derivationStrategy, DerivationFeature feature, KeyPath keyPath, NBXplorerNetwork network)
```
NicolasDorier: can you remove this ctor? it is almost the same as above. (maybe the same with different order)

Kukks:
@NicolasDorier  Rebased.

Is there anything left from your end because I think we are good to go with this PR

NBXplorer.Client/Models/TransactionResult.cs
```diff
@@ -34,7 +34,7 @@ public uint256 BlockId
 		}
 		public uint256 TransactionHash { get; set; }
 		Transaction _Transaction;
-		public Transaction Transaction
+		public virtual Transaction Transaction
```
NicolasDorier: useless, remove

NicolasDorier:
There is no test. You need at least to test all the routes that are used by btcpayserver. Also, you don't seem to expose the Asset Id in any callback, so it is impossible for btcpayserver to know what it received

Kukks:
> 
> 
> There is no test. You need at least to test all the routes that are used by btcpayserver (outisde of wallets). Also, you don't seem to expose the Asset Id in any callback, so it is impossible for btcpayserver to know what it received. Also test what happen if the utxo can't be unblinded.

NBX should not care about the assets transferred: this is handled in BTCPay.

Kukks:
@NicolasDorier  test added

Kukks:
Can we get this merged now @NicolasDorier ?

NicolasDorier:
Sorry, will work on it tomorrow and merge.

NBXplorer/Controllers/MainController.cs
```diff
@@ -614,7 +614,15 @@ private static TrackedSource GetTrackedSource(DerivationStrategyBase derivationS
...
 
-					txInfo.BalanceChange = txInfo.Outputs.Select(o => o.Value).Sum() - txInfo.Inputs.Select(o => o.Value).Sum();
+					if (network.NBitcoinNetwork.NetworkSet == NBitcoin.Altcoins.Liquid.Instance)
+					{
+						txInfo.BalanceChange = new MoneyBag(txInfo.Outputs.Select(o => o.Value).OfType<AssetMoney>().ToArray())
+												- new MoneyBag(txInfo.Inputs.Select(o => o.Value).OfType<AssetMoney>().ToArray());
+					}
+					else
+					{
+						txInfo.BalanceChange = txInfo.Outputs.Select(o => o.Value).OfType<Money>().Sum() - txInfo.Inputs.Select(o => o.Value).OfType<Money>().Sum();
```
Kukks: this wont blow up on blinded outputs?

NBXplorer/LiquidRepository.cs
```diff
@@ -0,0 +1,223 @@
...
+			}
+		}
+
+		protected override async Task AfterMatch(TrackedTransaction tx)
+		{
+			if (tx.TrackedSource is DerivationSchemeTrackedSource ts &&
+				tx.Transaction is ElementsTransaction elementsTransaction &&
+				tx is ElementsTrackedTransaction elementsTracked)
+			{
+				var unblinded = await _rpcClient.UnblindTransaction(
```
Kukks: this imports the blindingkey in elementsd

docs/API.md
```diff
@@ -1121,3 +1125,33 @@ A endpoint that can be used without the need for [authentication](#auth) which w
...
+}
+```
+
+The blinding key of the confidential address is derived directly from the `derivationScheme`.
+If the `scriptPubKey` `0/2` is generated, the blinding private key used by NBXplorer is the SHA256 of the scriptPubKey at `0/2/0`.
+
+In order to send in and out of liquid, we advise you to rely on the RPC command line interface of the liquid deamon.
+For doing this you need to [Generate a wallet](#wallet) with `importAddressToRPC` and `savePrivateKeys` set to `true`.
+
+Be careful to not expose your NBXplorer server on internet, your private keys can be [retrieved trivially](#getmetadata).
```
kallewoof: This sounds like you are revealing the wallet privkeys to anyone who is able to connect to your nbxplorer instance, which I assume is not the case. You're talking about people getting the xpubkey and deriving the blinding privkeys, right? Maybe "the private blinding keys, used to reveal the asset and value types of confidential transactions, can be [retrieved trivially](#getmetadata)"?

docs/API.md
```diff
@@ -1121,3 +1125,33 @@ A endpoint that can be used without the need for [authentication](#auth) which w
...
+* [Update Partially Signed Bitcoin Transaction](#updatepsbt) is not supported
+* [Scan UTXO Set](#scanUtxoSet) is not supported.
+* Any sort of recovery is not supported.
+
+The `AssetMoney` JSON format is:
+
+```json
+{
+	"assetId": "abc",
+	"value": 123
```
kallewoof: How do I get the asset and amount blinding factors from this?

How about the case where I have a custom blinding key (e.g. I make my own addresses and do not trust NBXplorer's blinding key derivation scheme), i.e. where NBXplorer cannot unblind the output?

docs/API.md
Kukks: Yes, NBX is not meant to be exposed publicly.   This does indeed allow anyone with access to your NBX instance to retrieve your private keys, when setting  `savePrivateKeys` & `importAddressToRPC`.

The main use-case for this generate wallet endpoint is to be able to create a wallet through NBX( generate a seed), track it, and be able to import to the Bitcoin/Elements/Shitcoin Core wallet(with `importAddressToRPC`) and use it to view balances or spend( with `savePrivateKeys`).

In terms of Liquid, this is mandatory currently, to not complicate the NBX codebase in terms of additional derivation scheme parsing options( which  there is a branch I have from before but added quite a lot of code for this PR)

docs/API.md
Kukks: if a tx is made to an address tracked by NBX, that is blinded by a key not available to NBX, it will be stored but its `Value`  property will be null.
I had a branch before where it would ignore any incoming txs where it could not unblind the tx.

NicolasDorier: @kallewoof if you have your own derivation stuff, NBXplorer can't decrypt but you will get the outputs anyway with the blinded transaction, just not the Value.

kallewoof: Yes, I realized after writing that you actually can't give me the blinding factors unless you know the blinding privkey. For the cases where you do, though, it would be helpful to include the blinding factors, as I could just pass those on when building a tx interactively with other people.

PR #202, opened 2019-11-15
Fix Health Check

junderw:
The results were always Degraded.

The logic was flawed, since anything & false is always false... and ok starts at false.


NicolasDorier:
I hate booleans

PR #205, opened 2019-11-25
Add import wallet API

Kukks:
This is almost identical to `generate` but allows you to use an existing seed you have.


NicolasDorier:
@Kukks can you instead just add `ExistingMnemonic` to `GenerateRequest`? This will prevent lot's of refactoring, review and duplication of doc.

Kukks:
Updated

NicolasDorier:
@Kukks  I will remove the MnemonicConverter, because we should check the Language.

PR #216, opened 2019-12-06
save passphrase meta too

Kukks:


NicolasDorier:
why do you need passphrase? You have `MasterHDKey`

Kukks:
ah true

PR #218, opened 2019-12-07
refactor deriv strategy options and allow elements unblinded flag

Kukks:
This lets all the derivation implementations have access to all options + creates a generic way to read additional options along with generating the string suffixes.

Then I leveraged it and added support for `-[unblinded]` for liquid without needing to creating derived classes fro derivations.

### DIFF
--- a/NBXplorer.Client/DerivationStrategy/DerivationStrategy.cs
+++ b/NBXplorer.Client/DerivationStrategy/DerivationStrategy.cs
@@ -12,6 +12,7 @@ namespace NBXplorer.DerivationStrategy
 	public class DerivationStrategyOptions
 	{
 		public ScriptPubKeyType ScriptPubKeyType { get; set; }
+		public Dictionary<string,bool> AdditionalOptions { get; set; } = new Dictionary<string, bool>();
 
 		/// <summary>
 		/// If true, in case of multisig, do not reorder the public keys of an address lexicographically (default: false)
@@ -40,7 +41,6 @@ namespace NBXplorer.DerivationStrategy
 		}
 
 		readonly Regex MultiSigRegex = new Regex("^([0-9]{1,2})-of(-[A-Za-z0-9]+)+$");
-		static DirectDerivationStrategy DummyPubKey = new DirectDerivationStrategy(new ExtKey().Neuter().GetWif(Network.RegTest)) { Segwit = false };
 		public DerivationStrategyBase Parse(string str)
 		{
 			var strategy = ParseCore(str);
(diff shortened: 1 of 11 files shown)


NBXplorer.Client/NBXplorerNetworkProvider.Liquid.cs
```diff
@@ -29,6 +29,11 @@ public override BitcoinAddress CreateAddress(DerivationStrategyBase derivationSt
 
 			public static Key GenerateBlindingKey(DerivationStrategyBase derivationStrategy, KeyPath keyPath)
```
NicolasDorier: this should throw `InvalidOperationException` if it is unblinded.

NBXplorer/Altcoins/Liquid/LiquidRepository.cs
```diff
@@ -192,7 +190,7 @@ protected override async Task AfterMatch(TrackedTransaction tx, IReadOnlyCollect
 					.Select(kv => (KeyPath: kv.KeyPath,
 								   Address: kv.Address as BitcoinBlindedAddress,
 								   BlindingKey: NBXplorerNetworkProvider.LiquidNBXplorerNetwork.GenerateBlindingKey(ts.DerivationStrategy, kv.KeyPath)))
-					.Where(o => o.Address != null)
+					.Where(o => o.Address != null && o.BlindingKey != null)
```
NicolasDorier: the whole aftermatch should be skipped if unblinded, no need of hacking.

NBXplorer.Client/DerivationStrategy/DerivationStrategy.cs
```diff
@@ -11,14 +11,52 @@ namespace NBXplorer.DerivationStrategy
 {
 	public class DerivationStrategyOptions
 	{
-		public ScriptPubKeyType ScriptPubKeyType { get; set; }
+
+		public ScriptPubKeyType ScriptPubKeyType
```
NicolasDorier: Don't do this, this is horrible to calculate that everytimes instead of once and affecting ScritPuBKeyType

Kukks: fixed

NBXplorer/Altcoins/Liquid/LiquidRepository.cs
Kukks: fixed

NBXplorer.Client/NBXplorerNetworkProvider.Liquid.cs
Kukks: fixed

Kukks:
Not sure why this fails, any idea?

NBXplorer.Client/DerivationStrategy/DerivationStrategy.cs
NicolasDorier: this is not fixed, it should not be calculated.

NicolasDorier:
I don't like the refactoring on AdditionalOptions. This make the code worse for a very marginal feature.
I suggest you to make a derived class for the `DervationSchemeFactory` where the elements one extract `unblinded`. You can use `AdditionalOptions` to store it, but keep the rest outside of AdditionalOptions.

PR #225, opened 2019-12-17
Add MinUtxoValue for ignoring spam

### LINKED ISSUE #186: Feature Request: Ignore small amounts (configurable)
Tracking and scanutxoset should remove UTXOs below a specified threshold.

Default is none.

Could be an environment/runtime parameter.

Use case: Do not want to accidentally spend those spammy 1 satoshi inputs or recognize utxos that are below minimum deposit limit.

This is a very common metric for keeping spam from clogging up our systems.

junderw:
Fixes #186 (Maybe??? lol)

Am I heading in the right direction here?

### DIFF
--- a/NBXplorer/Configuration/ExplorerConfiguration.cs
+++ b/NBXplorer/Configuration/ExplorerConfiguration.cs
@@ -140,7 +140,7 @@ namespace NBXplorer.Configuration
 					}
 
 					chainConfiguration.StartHeight = config.GetOrDefault<int>($"{network.CryptoCode}.startheight", -1);
-					chainConfiguration.MinUtxoValue = config.GetOrDefault<Money>($"{network.CryptoCode}.minutxovalue", Money.Satoshis(1));
+					chainConfiguration.MinUtxoValue = Money.Satoshis(config.GetOrDefault<int>($"{network.CryptoCode}.minutxovalue", 1));
 					chainConfiguration.HasTxIndex = config.GetOrDefault<bool>($"{network.CryptoCode}.hastxindex", false);
 
 					ChainConfigurations.Add(chainConfiguration);
--- a/NBXplorer/Configuration/ExplorerConfiguration.cs
+++ b/NBXplorer/Configuration/ExplorerConfiguration.cs
@@ -37,7 +37,7 @@ namespace NBXplorer.Configuration
 			get;
 			internal set;
 		}
-		public int MinUtxoValue
+		public Money MinUtxoValue
 		{
 			get;
 			internal set;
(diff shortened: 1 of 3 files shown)


junderw:
Will this also block spam unconfirmed transactions as well? (I noticed it was talking about blockId)

junderw:
I should also add stuff to SCANUtxoSetService and LiquidRepository

junderw:
I should modify `SaveMatches` instead of modifying SCANUtxo. SCANUtxo can report 200 matches found and if 50 of them are spam, there will only be 150 added to the repo.

junderw:
actually... SCANUtxo is too much of a pain... I think this PR is fine as is.

junderw:
`I think I actually need to modify `UpdateRepository` in SCANUtxoSetService... so I will give up.

NBXplorer/Repository.cs
```diff
@@ -735,6 +736,10 @@ public int MaxPoolSize
 		{
 			get; set;
 		} = 30;
+		public int MinUtxoValue
```
NicolasDorier: Change `int` to `Money`.

NBXplorer/Configuration/ExplorerConfiguration.cs
```diff
@@ -135,6 +140,7 @@ public ExplorerConfiguration LoadArgs(IConfiguration config)
 					}
 
 					chainConfiguration.StartHeight = config.GetOrDefault<int>($"{network.CryptoCode}.startheight", -1);
+					chainConfiguration.MinUtxoValue = config.GetOrDefault<int>($"{network.CryptoCode}.minutxovalue", 1);
```
NicolasDorier: You can convert `int` to `Money` with `Money.Satoshis(value)`

junderw:
Please review the new commit. Thanks.

junderw:
Question: Would the following also fix ScanUtxoSet???

```diff
diff --git a/NBXplorer/ScanUTXOSetService.cs b/NBXplorer/ScanUTXOSetService.cs
index bf87d26..52206ac 100644
--- a/NBXplorer/ScanUTXOSetService.cs
+++ b/NBXplorer/ScanUTXOSetService.cs
@@ -249,7 +249,7 @@ namespace NBXplorer
                                                          BlockId: blockIdsByHeight.TryGet(o.First().Height),
                                                          TxId: o.Select(c => c.Coin.Outpoint.Hash).FirstOrDefault(),
                                                          KeyPathInformations: o.Select(c => scannedItems.KeyPathInformations[c.Coin.ScriptPubKey]).ToList()))
-                               .Where(o => o.BlockId != null)
+                               .Where(o => o.BlockId != null && o.Coin.Amount.Satoshis >= repo.MinUtxoValue)
                                .Select(o =>
                                {
                                        foreach (var keyInfo in o.KeyPathInformations)
```

junderw:
Also...

I noticed that this PR and the ScanUTXOSet diff I make up there seems to talk about blocks...

Does this merged PR also filter out unconfirmed UTXOs by value as well? Or would I need to fix something else?

PR #230, opened 2019-12-30
Add unblinded flag , alternative edition

Kukks:


NBXplorer/LiquidRepository.cs
```diff
@@ -206,6 +208,30 @@ protected override async Task AfterMatch(TrackedTransaction tx, IReadOnlyCollect
 			}
 		}
 
+		
+		public override async Task<KeyPathInformation> GetUnused(DerivationStrategyBase strategy, DerivationFeature derivationFeature, int n, bool reserve)
```
NicolasDorier: This hack will probably have side effect which will be worse than just considering blinded and unblinded as different wallet.

I don't understand why you need to do this. No user will want one store in blinded and the other unblinded and share the same indexes.

Kukks:
The issue here is that if you are tracking the same xpub with blinded and
unblinded options, they will generate the same script pub when requesting
an unused address. Once a payment is made to the blinded address, it will
trigger the unblinded tracker too.

If you have two stores on btcpay, each using the same xpub:
Store A: uses xpub - [unblinded
Store B: uses xpub
Generate invoice on both stores.
Both invoices have the same script pub key, but different addresses( store
b is the blinded address format, but its unblinded address is the same)

If you pay Store A s invoice, the invoice from store B is also paid.
If you pay Store B s invoice,the invoice from store A is not paid as it is
blinded

NicolasDorier:
Good point. Though I am not convinced it is a problem worth solving, as it means that a user need:

1. Use lightning
2. Have two stores
3. Configure both stores with same xpub but and one blinded/other unblinded

NBXplorer/LiquidRepository.cs
Kukks: It's more for the case that a user starts off with one option and then switches to the other, only to find out that there is address reuse happening and that some calls to GetUnused would actually return indexes that have been used. 

Maybe I am overthinking this?

Kukks:
> 
> 
> Good point. Though I am not convinced it is a problem worth solving, as it means that a user need:
> 
>     1. Use lightning
> 
>     2. Have two stores
> 
>     3. Configure both stores with same xpub but and one blinded/other unblinded
> 
> 
> Not really something that should happen everyday...

Why do you need to use lightning?

NicolasDorier:
I really don't like this fix though.
I can't convince myself there is not side effect which will waste our time in debugging later. And it is not easy to understand: Why would different label sometimes map to the same addresses and sometime not?

This is also for fixing something that will happen only one time for a very limited set of people.

Kukks:
Fair enough, will remove the hack. I guess a documented solution would be
to Rescan if you encounter the issue

Kukks:
updated and removed the workaround

NBXplorer.Client/DerivationStrategy/IStrategy.cs
```diff
@@ -38,11 +41,18 @@ public Derivation GetDerivation(KeyPath keyPath)
 		}
 		public abstract Derivation GetDerivation();
 
-		protected abstract string StringValue
+		protected abstract string StringValueCore
```
NicolasDorier: remove this refactoring, this is not needed anymore

NBXplorer/Extensions.cs
```diff
@@ -48,7 +48,7 @@ internal static void AddRange<T>(this HashSet<T> hashset, IEnumerable<T> element
 		}
 		internal static uint160 GetHash(this DerivationStrategyBase derivation)
 		{
-			var data = Encoding.UTF8.GetBytes(derivation.ToString());
+			var data = Encoding.UTF8.GetBytes(derivation.StringValue);
```
NicolasDorier: remove this

NBXplorer.Client/DerivationStrategy/IStrategy.cs
Kukks: This one is needed in order to not make each derivation type have to do `StringValue => $"{StringValueCore}{GetSuffixOptionsString()}";`

NicolasDorier:
You need to add tests to make sure the that the derivation schemes are insensitive to the order and the casing of the labels. Else we will end up with various issues.

Kukks:
> 
> 
> You need to add tests to make sure the that the derivation schemes are insensitive to the order and the casing of the labels. Else we will end up with various issues. (only need tests at the factory level)

Done

PR #241, opened 2020-02-12
enable to specify polling interval as configuration value

joemphilips:


junderw:
LGTM

NicolasDorier:
I can merge that, but I would like to understand the reason.
This poll should normally not affect you at all. This is a poll which check if the bitcoind is still working.

NBXplorer/Configuration/DefaultConfiguration.cs
```diff
@@ -44,6 +44,7 @@ protected override CommandLineApplication CreateCommandLineApplicationCore()
 				app.Option($"--{crypto}minutxovalue", $"The minimum value of tracked UTXOs, any UTXO with value less than this is ignored. (default: 1 (satoshi))", CommandOptionType.SingleValue);
 				app.Option($"--{crypto}nodeendpoint", $"The p2p connection to a Bitcoin node, make sure you are whitelisted (default: default p2p node on localhost, depends on network)", CommandOptionType.SingleValue);
 				app.Option($"--{crypto}hastxindex", "If true, NBXplorer will try to fetch missing transactions from the local node (default: false)", CommandOptionType.BoolValue);
+				app.Option($"--{crypto}pollinginterval", $"Polling interval (in seconds) for querying about UTXO to the chain. (default: 60 seconds)", CommandOptionType.SingleValue);
```
NicolasDorier: it does not query UTXOs, it just connect to the bitcoind or do nothing if already connected.

junderw:
NBX uses ZMQ for new transactions and blocks, right?

joemphilips:
> I think you are mistaken on what you think this code does. There is no UTXO query here, this is just a loop to control the connection to bitcoind.

Yep, I was misunderstanding. 

Well, we were facing some weird error when we were trying to send tx in a row. I first thought it is because there is a mismatch in utxo-set between nbxplorer and bitcoind in some situation. thus making the polling interval short well mitigate the problem.


umm.. I'd better first make clear what is the exact cause.

NicolasDorier:
@junderw no, it uses P2P.

NicolasDorier:
@joemphilips NBXplorer is directly connected to BitcoinD via P2P to get blocks and transactions.

I think the problem was that due to fee spike, the transaction was not gossiped for some reason. (maybe too low fee? or evicted from mempool? I am unsure)

The other problem which may have happened, but that I really doubt is that there is a global lock on the database. In case you query a huge wallet, it fetch all the transactions which may take time. Meanwhile, the transaction is waiting for its turned to be saved in the DB.

If this was the problem, you would experience huge delay when you create a PSBT or list transactions/utxos. If you ever experience this, there is a call in the API to manually prune transactions whose all UTXOs has been spent for a while.

PR #242, opened 2020-02-12
Run SaveTransactions and GetMatches/SaveMatches whenever broadcast succeeds.

junderw:
Let me know if this is a good way to go about it.

### DIFF
--- a/NBXplorer/Controllers/MainController.cs
+++ b/NBXplorer/Controllers/MainController.cs
@@ -943,6 +943,7 @@ namespace NBXplorer.Controllers
 					};
 				}
 				await waiter.RPC.SendRawTransactionAsync(tx);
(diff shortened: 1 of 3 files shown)


NBXplorer/Controllers/MainController.cs
```diff
@@ -969,6 +970,7 @@ private async Task<AnnotatedTransactionCollection> GetAnnotatedTransactions(Repo
 					{
 						await waiter.RPC.SendRawTransactionAsync(tx);
 						Logs.Explorer.LogInformation($"{network.CryptoCode}: Broadcast success");
+						await waiter.GetExplorerBehavior().SaveMatches(tx);
```
NicolasDorier: Should be `await waiter.GetExplorerBehavior()?.SaveMatches(tx);` but I can fix.

PR #244, opened 2020-02-20
Custom auth header

KeyFiDennis:
Enable passing of custom authentication headers to the client. If NBX is hidden behind a proxy that provides authentication capabilities the cookie authentication is not sufficient. Passing a custom Auth Header allows for different authentication methods via Token, Cookie, HttpBasic, OAuth....

The changes are minimal but enable some customization for users if needed.

### DIFF
--- a/NBXplorer.Client/ExplorerClient.cs
+++ b/NBXplorer.Client/ExplorerClient.cs
@@ -20,7 +20,7 @@ namespace NBXplorer
 {
 	public class ExplorerClient
 	{
-		internal interface IAuth
+		public interface IAuth
 		{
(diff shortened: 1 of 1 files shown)


NicolasDorier:
Hey thanks for the contribution.
What about we just expose the IAuth interface and classes as public then allow to inject it via the constructor? You would then be able to use your own implementation of the interface just fine.

KeyFiDennis:
That would be probably an even better Solution that makes it customizable.

KeyFiDennis:
@NicolasDorier exposed it as you suggested and added to the constructor.

PR #245, opened 2020-02-21
Update Dbriize

joemphilips:
**Please don't merge this**

I'm just checking how tests run in CI environment.

To explain the purpose, I am trying to bump the version of DBriize according to the DBreeze.
In the hope of solving the problem we are facing. (So this PR should be moved to a DBriize repository after I have checked it is working correctly.)


joemphilips:
checked it is working correctly and tests pass.
I have created PR here https://github.com/NicolasDorier/DBriize/pull/2

PR #247, opened 2020-03-06
Add RPC Client Proxy

Kukks:
This let's you use nbx as a direct proxy to the node rpc.


Kukks:
@NicolasDorier removed all the unneeded logic. Should be good to go

NicolasDorier:
It should not be on by default, there is a security implication.
Can you add a global configuration switch to enable this feature and default to false.

Kukks:
@NicolasDorier  done

Kukks:
oops did not mean to close

PR #257, opened 2020-05-13
Add server alias option

junderw:
Adding it to HTTP response was too hard for me. (Microsoft documentation is a jungle)

If you can easily change this to be a HTTP response header for all endpoints that should work too.

### DIFF
--- a/NBXplorer.Client/Models/StatusResult.cs
+++ b/NBXplorer.Client/Models/StatusResult.cs
@@ -69,6 +69,11 @@ namespace NBXplorer.Models
 			get;
 			set;
 		}
+		public string? Alias
+		{
+			get;
+			set;
+		}
 		[JsonConverter(typeof(Newtonsoft.Json.Converters.StringEnumConverter))]
 		public NetworkType NetworkType
 		{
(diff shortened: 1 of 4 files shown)


NicolasDorier:
what should be the header name?

junderw:
Two things:

1. Response to your question.
2. Suggestion for default value.

-------

According to this answer: (https://stackoverflow.com/a/3561399/3074620) it used to be recommended to add X- before custom headers, but now it is recommended to NOT use X-...

So we should not use X- I think. But some people still argue X- is better

`Nbxplorer-Alias` sounds good.

-------

Also, instead of default being null, I think default should be some random generated UUID

junderw:
The argument of "if this header will never become a worldwide standard, add X-" sounds reasonable... so maybe `X-Nbxplorer-Alias` is better. aggghhh.

Standards are to hard. lol

NicolasDorier:
`X-` is actually deprecated https://tools.ietf.org/html/rfc6648

NicolasDorier:
will use `instance-name`

NicolasDorier:
Done on https://github.com/dgarage/NBXplorer/commit/0e316fba7b003520501ccbe720ed7d088efc5983

PR #262, opened 2020-05-23
[WIP] Randomize fingerprint of transactions

NicolasDorier:
We will randomize fingerprint of transactions created by NBXplorer by maintaining a distribution of the fingerprint over the last 5 blocks.

Related to https://github.com/zkSNACKs/WalletWasabi/issues/3625


NicolasDorier:
ping @lontivero I think you will be interested by this code. Especially the `FingerprintDistribution` class.
Very easy to use, you feed a block and it create the distribution. Then you can make query over it.

For example, if you know your user is using P2WPKH and that he want RBF, you can readjust the distribution based on this fact to get a proper distribution of other parameters such as FeeSnipping, or LowR etc...

PR #266, opened 2020-06-06
Allow attempting to set the full tx for segwit inputs

Kukks:
An new option to allow *attempting* to set `non_witness_utxo` for segwit inputs on PSBT create/update.


NicolasDorier:
Test fail.

Can you rename to `alwaysIncludeNonWitnessUXTO`?

Kukks:
renamed and test fixed

PR #268, opened 2020-06-11
Allow GetEvents to fetch limit events from latest

junderw:
I first separated into 2 different endpoints in commit d2c0a7a but then decided to just add a getLatest parameter.

This is useful for when you might have tons of events, and your client lost its state, but you decide "if we lose state on the client, we should only care about the most recent X events."

Currently, you must get all events to find the last event id.

Alternative solution:

Add new endpoint that will get the latest eventID from the server.

### DIFF
--- a/NBXplorer.Client/LongPollingNotificationSession.cs
+++ b/NBXplorer.Client/LongPollingNotificationSession.cs
@@ -84,5 +84,24 @@ namespace NBXplorer
 					  .ToArray();
 			return evtsObj;
 		}
+
+		public NewEventBase[] GetLatestEvents(int limit = 10, CancellationToken cancellation = default)
+		{
+			return GetLatestEventsAsync(limit, cancellation).GetAwaiter().GetResult();
+		}
+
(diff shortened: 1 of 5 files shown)


junderw:
Using separate endpoint as per Twitter convo.

NBXplorer.Client/LongPollingNotificationSession.cs
```diff
@@ -84,5 +84,26 @@ public async Task<NewEventBase[]> GetEventsAsync(long lastEventId = 0, int? limi
 					  .ToArray();
 			return evtsObj;
 		}
+
+		public NewEventBase[] GetLatestEvents(int limit = 10, bool longPolling = false, CancellationToken cancellation = default)
```
NicolasDorier, line 88: remove long polling

NicolasDorier:
tests do not pass

junderw:
SelectFrom gets the events AFTER the arg you pass it (not including the number)

All is fixed now.

Thanks.

PR #277, opened 2020-06-23
Use NBitcoin level CryptoCode for GRS

gruve-p:
Not sure if this is correct or should not be merged.
I am trying to debug importing an ypub and zpub issue for GRS on https://www.grspay.com and saw this missing. Not sure if this related to the issue I am having.

Currently I can only succesfully import xpub for GRS (mainnet and testnet). When importing an ypub and zpub I get the error:  `Invalid Derivation Scheme`
When I convert the ypub and zpub to xpub (by changing the version byte) through: https://groestlcoin.org/xpub-converter/ then it recognises the bech32 address and works.


gruve-p:
The issue I was having is fixed through https://github.com/btcpayserver/btcpayserver/pull/1688

NicolasDorier:
yeah not needed.

PR #294, opened 2020-09-16
[Add] DogeCash Support

Liquid369:
This PR builds successfully and I have not gotten any issues during the test cases.

### DIFF
--- a/NBXplorer.Client/NBXplorerNetworkProvider.DogeCash.cs
+++ b/NBXplorer.Client/NBXplorerNetworkProvider.DogeCash.cs
@@ -1,0 +1,24 @@
+using NBitcoin;
+using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace NBXplorer
+{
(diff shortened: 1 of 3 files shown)


NicolasDorier:
I don't think it can work, if I remember DogeCash is not supported by NBitcoin anymore because it was breaking NBitcoin.Altcoins for some reason and nobody fixed it.

Liquid369:
There was a recent PR https://github.com/MetacoSA/NBitcoin/pull/792
Fixing invalid hexstring error and is still supported.

Then submitted a PR yesterday as well to update versions
https://github.com/MetacoSA/NBitcoin/pull/918

Liquid369:
If there is some errors still I will recheck through today

NicolasDorier:
as you can see dogecash crash it all

NicolasDorier:
maybe related, maybe not. NBitcoin's AltcoinNetworkSets.cs Dogecash is commented. It was commented because it was crashing in similar way.

PR #297, opened 2020-11-22
add GetBalance by address methods

CodeFunta:


NBXplorer.Client/ExplorerClient.cs
```diff
@@ -313,7 +313,7 @@ public void CancelReservation(DerivationStrategyBase strategy, KeyPath[] keyPath
 			CancelReservationAsync(strategy, keyPaths, cancellation).GetAwaiter().GetResult();
 		}
 
-		public GetBalanceResponse GetBalance(DerivationStrategyBase userDerivationScheme, CancellationToken cancellation = default)
```
NicolasDorier, line 316: the cancellation token should be passed to `GetBalanceAsync`, not removed.

NBXplorer.Client/ExplorerClient.cs
```diff
@@ -322,6 +322,16 @@ public Task<GetBalanceResponse> GetBalanceAsync(DerivationStrategyBase userDeriv
 			return SendAsync<GetBalanceResponse>(HttpMethod.Get, null, "v1/cryptos/{0}/derivations/{1}/balance", new[] { CryptoCode, userDerivationScheme.ToString() }, cancellation);
 		}
 
+
+		public GetBalanceResponse GetBalance(string address)
```
NicolasDorier, line 326: Should be a strongly typed `BitcoinAddress`, not a string.

NBXplorer.Client/ExplorerClient.cs
```diff
@@ -322,6 +322,16 @@ public Task<GetBalanceResponse> GetBalanceAsync(DerivationStrategyBase userDeriv
...
+		{
+			return GetBalanceAsync(address).GetAwaiter().GetResult();
+		}
+		public Task<GetBalanceResponse> GetBalanceAsync(string address, CancellationToken cancellation = default)
+		{
+			return SendAsync<GetBalanceResponse>(HttpMethod.Get, null, "v1/cryptos/{0}/addresses/{1}/balance", new[] { CryptoCode, address }, cancellation);
```
NicolasDorier, line 332: Reuse `GetBalanceAsync(DerivationStrategyBase userDerivationScheme)` instead, there is a subclass to pass address only.

NBXplorer.Client/ExplorerClient.cs
CodeFunta: Ok, But GetBalance is sync version and cancelationToken don't used inside GetBalanceAsync  call. I'll change it to use

NicolasDorier: this is a bug, we should use it in `GetBalanceAsync `

NBXplorer.Client/ExplorerClient.cs
CodeFunta: Ok

NBXplorer.Client/ExplorerClient.cs
CodeFunta: I've only seen logic with TrackedSource when address was used. Still don't understand how to create DerivationStrategy from factory for btc address

NBXplorer.Client/ExplorerClient.cs
CodeFunta: fix was pushed

NBXplorer.Client/ExplorerClient.cs
CodeFunta: Changed with new commit

NBXplorer.Client/ExplorerClient.cs
NicolasDorier: Ah you are right this is a tracked source, not a derivation strategy.

PR #301, opened 2020-12-27
Bump NBitcoin

Kukks:
Bumps NBitcoin and fixes some breaking changes


NicolasDorier:
conflits

NicolasDorier:
I think we don't need it anymore, as I bumped it.

PR #305, opened 2021-02-10
add support for the altcoin Althash

denuoweb:
thank you for providing this as open source


denuoweb:
I will have to wait until NBitcoin new version comes out before I do a PR here.

Issue #321, opened 2021-07-29
Immature coinbase UTXOs counted towards confirmed balance


sageprogrammer:
**Description**
Currently the GetBalance function does not differetiate between immature coinbase txn and confirmed txn.

**Expected**
GetBalance function must return 3 fields, confirmed, immature & unconfirmed

**Reality**
GetBalance function only returns 2 fields, confirmed & unconfirmed

https://github.com/dgarage/NBXplorer/blob/853955171e8c19bffee47831a30fb4f8dc442cce/NBXplorer/Controllers/MainController.cs#L838

sageprogrammer:
I am working on this

PR #323, opened 2021-07-30
Support for Immature Coin Base Transaction

sageprogrammer:
Closes: #321

### DIFF
--- a/docker-compose.regtest.yml
+++ b/docker-compose.regtest.yml
@@ -54,10 +54,6 @@ services:
   rabbitmq:
     image: rabbitmq:3-management
     hostname: rabbitmq
-    environment:
-      "RABBITMQ_DEFAULT_PASS": "guest"
-      "RABBITMQ_DEFAULT_USER": "guest"
-      "RABBITMQ_DEFAULT_VHOST": "/"
     ports:
       - 4369:4369
       - 5671:5671
--- a/NBXplorer.Client/Models/GetBalanceResponse.cs
+++ b/NBXplorer.Client/Models/GetBalanceResponse.cs
@@ -9,6 +9,7 @@ namespace NBXplorer.Models
 	{
 		public IMoney Unconfirmed { get; set; }
 		public IMoney Confirmed { get; set; }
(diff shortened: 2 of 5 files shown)


sageprogrammer:
closes #321

NicolasDorier:
It seems like a good solution. I a wondering if `Total` should take into account immature transactions.

NicolasDorier:
Can you update the doc in `docs/API.md`?

NicolasDorier:
Looking at the code, the Total already should not report immature coins. So something else might be wrong.

docker-compose.regtest.yml
```diff
@@ -54,10 +54,6 @@ services:
   rabbitmq:
     image: rabbitmq:3-management
     hostname: rabbitmq
-    environment:
```
NicolasDorier, line 57: why rmoving this?

NicolasDorier:
Actually, I manage to reproduce the immature issue. Trying to understand why my code did not work, I thought taking care of it.

NicolasDorier:
I will slightly change the API, but fix is good thanks a lot!

docker-compose.regtest.yml
sageprogrammer: Actually when I tried to run rabbitmq via docker-compose, it was complaining that 
rabbitmq_1          | error: RABBITMQ_DEFAULT_PASS is set but deprecated
rabbitmq_1          | error: RABBITMQ_DEFAULT_USER is set but deprecated
rabbitmq_1          | error: RABBITMQ_DEFAULT_VHOST is set but deprecated
rabbitmq_1          | error: deprecated environment variables detected
rabbitmq_1          | 
rabbitmq_1          | Please use a configuration file instead; visit https://www.rabbitmq.com/configure.html to learn more


Also I found that even if we do not provide any default ones, according to https://www.rabbitmq.com/access-control.html#default-state, It still uses same values.

NicolasDorier:
tests and doc, and slight API change on https://github.com/dgarage/NBXplorer/pull/324 let me know.

PR #324, opened 2021-07-31
Document and properly test immature transaction handling

NicolasDorier:
https://github.com/dgarage/NBXplorer/issues/321
Ping @sageprogrammer

What do you think of this API?

Technically, immature UTXO and transactions are "Confirmed".
That said, in 99.9999% of cases, what people want when they say "confirmed" they mean "funds we can spend".

I thought including immature transactions in confirmed transactions. But then it would have been confusing: Why the confirmed utxo doesn't include immature utxos, but confirmed transactions does?

### DIFF
--- a/NBXplorer.Tests/UnitTest1.cs
+++ b/NBXplorer.Tests/UnitTest1.cs
@@ -2663,6 +2663,45 @@ namespace NBXplorer.Tests
 				tester.Notifications.WaitForBlocks(blockId);
 				var savedTx = tester.Client.GetTransaction(txId);
 				Assert.Equal(blockId[0], savedTx.BlockId);
+
+				// Ensure the current state is correct
+				var balance = tester.Client.GetBalance(pubkey);
+				Assert.Equal(Money.Coins(2.5m), balance.Confirmed);
+				Assert.Equal(Money.Coins(0.0m), balance.Unconfirmed);
+				Assert.Equal(Money.Coins(2.5m), balance.Total);
+				Assert.Equal(Money.Coins(0.0m), balance.Immature);
+				Assert.Equal(Money.Coins(2.5m), balance.Available);
+				Logs.Tester.LogInformation("Let's mine, and check that the balance is not updated until maturity occurs");
+				var blkid = tester.RPC.GenerateToAddress(1, tester.AddressOf(key, "0/0"))[0];
+				var blk = tester.RPC.GetBlock(blkid);
+				var minedTxId = blk.Transactions[0].GetHash();
+				tester.Notifications.WaitForTransaction(pubkey, minedTxId);
+				balance = tester.Client.GetBalance(pubkey);
+				Assert.Equal(Money.Coins(52.5m), balance.Confirmed);
+				Assert.Equal(Money.Coins(0.0m), balance.Unconfirmed);
+				Assert.Equal(Money.Coins(52.5m), balance.Total);
+				Assert.Equal(Money.Coins(50.0m), balance.Immature);
+				Assert.Equal(Money.Coins(2.5m), balance.Available);
+				var utxos = tester.Client.GetUTXOs(pubkey);
+				Assert.DoesNotContain(utxos.Confirmed.UTXOs, u => u.Outpoint.Hash == minedTxId);
+				Assert.DoesNotContain(utxos.Unconfirmed.UTXOs, u => u.Outpoint.Hash == minedTxId);
+				var transactions = tester.Client.GetTransactions(pubkey);
+				Assert.Contains(transactions.ConfirmedTransactions.Transactions, u => u.TransactionId == minedTxId);
+				Assert.Contains(transactions.ImmatureTransactions.Transactions, u => u.TransactionId == minedTxId);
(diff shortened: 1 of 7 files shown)


NBXplorer.Tests/UnitTest1.cs
```diff
@@ -2663,6 +2663,21 @@ public void CanTrack()
...
+				var minedTxId = blk.Transactions[0].GetHash();
+				tester.Notifications.WaitForTransaction(pubkey, minedTxId);
+				var balanceAfter = tester.Client.GetBalance(pubkey);
+				Assert.Equal(Money.Coins(50m), (Money)balanceAfter.Total - (Money)balanceAfter.Available);
+				var utxos = tester.Client.GetUTXOs(pubkey);
+				Assert.DoesNotContain(utxos.Confirmed.UTXOs, u => u.Outpoint.Hash == minedTxId);
+				Assert.DoesNotContain(utxos.Unconfirmed.UTXOs, u => u.Outpoint.Hash == minedTxId);
+				var transactions = tester.Client.GetTransactions(pubkey);
+				Assert.DoesNotContain(transactions.ConfirmedTransactions.Transactions, u => u.TransactionId == minedTxId);
+				Assert.Contains(transactions.ImmatureTransactions.Transactions, u => u.TransactionId == minedTxId);
```
sageprogrammer, line 2680: I think we could add another part to the test, that when the coinbase txn do get matured they are included in confirmed txn list.

NicolasDorier: Done

sageprogrammer:
I think API change is good, as the one using it ( For eg btcpayserver) can easily infer that "available balance" is the one that contains the bitcoins that are spendable (maybe after some confirmations) in contrast to the "total balance" that may contain some that require many confirmations( i.e 100)

NBXplorer/AnnotatedTransactionCollection.cs
```diff
@@ -204,9 +204,8 @@ public AnnotatedTransactionCollection(ICollection<TrackedTransaction> transactio
 			{
 				if (tx.Height is int)
 				{
-					if(tx.IsMature)
-						ConfirmedTransactions.Add(tx);
-					else
+					ConfirmedTransactions.Add(tx);
+					if (!tx.IsMature)
 						ImmatureTransactions.Add(tx);
```
sageprogrammer, line 209: So now Confirmed transactions will contain Immature txns also?

NBXplorer/Controllers/MainController.cs
```diff
@@ -854,13 +854,14 @@ async Task<(uint256 BlockId, Transaction Transaction, DateTimeOffset BlockTime)>
...
+			var immature = CalculateBalance(network, transactions.ImmatureTransactions);
 			var balance = new GetBalanceResponse()
 			{
 				Confirmed = CalculateBalance(network, transactions.ConfirmedTransactions),
 				Unconfirmed = CalculateBalance(network, transactions.UnconfirmedTransactions),
-				Immature = CalculateBalance(network,transactions.ImmatureTransactions),
 			};
-			balance.Total = balance.Confirmed.Add(balance.Unconfirmed).Add(balance.Immature);
+			balance.Available = balance.Confirmed.Add(balance.Unconfirmed);
+			balance.Total = balance.Available.Add(immature);
```
sageprogrammer, line 864: I think this may lead to double lead to double counting of immature txns.

NicolasDorier: Yes, I rewrote the test to use hard coded number to make sure it was calculating right.

I also put back the field `immature`. It is clearer with it.

NicolasDorier:
Unconfirmed transactions are spendable.
I did some 180 degree turn on some stuff:

* `balance.Confirmed` include immature.
* Added back `balance.Immature` it is indeed clearer API with it.
* `transactions.Confirmed` include immature. (because the user expect to see the transaction, even if unspendable)
* `utxos.Confirmed` **doesn't** contains immature. The reason is that when you get confirmed UTXO, you likely want to build a transaction to spend something right away.

NBXplorer/AnnotatedTransactionCollection.cs
NicolasDorier: yes. I thought about how I am using ConfirmedTransaction right now when I use this API (for example in BTCPay).
In BTCPay I do `unconfirmedTransactions.Concat(confirmedTransactions)` to get the whole list. In this list, we expect to see the immature transactions. Imagine somebody call bitcoin core's `generatetoaddress` but doesn't see the transaction in the list. It would be surprising to him.

sageprogrammer: I see, thats a valid point.

NicolasDorier:
I pushed a NBXplorer.Client version 4.0.1 and 2.1.54. You can make a PR on BTCPayServer?

sageprogrammer:
I tried switching to  NBXplorer client to 4.0.x and git version but it caused some errors.
The problem is due to mismatch of Sign & SignAll function signatures.

PR #327, opened 2021-08-12
Get Getmatches ready for taproot

sageprogrammer:
Created a Draft `GetKeyInformations(OutPoint[] outpoints)` function that gets `KeyPathInforation` for each of the `Outpoint`s provided.
- In future this function will fetch from the DB.

Refactored the code of GetMatches function for taproot

### DIFF
--- a/NBXplorer/Repository.cs
+++ b/NBXplorer/Repository.cs
@@ -1,4 +1,4 @@
-﻿using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Logging;
 using System.Linq;
 using NBitcoin;
 using System;
@@ -175,7 +175,7 @@ namespace NBXplorer
 		private async ValueTask<DBTrieEngine> OpenEngine(string directory, CancellationToken cancellationToken)
 		{
 			int tried = 0;
-			retry:
+		retry:
 			try
 			{
 				return await DBTrie.DBTrieEngine.OpenFromFolder(directory);
@@ -407,6 +407,10 @@ namespace NBXplorer
 			return new Index(tx, $"{_Suffix}Scripts", $"{scriptPubKey.Hash}");
 		}
 
+		Index GetOutPointsIndex(DBTrie.Transaction tx, OutPoint outPoint)
+		{
+			return new Index(tx, $"{_Suffix}OutPoints", $"{outPoint}");
+		}
 		Index GetHighestPathIndex(DBTrie.Transaction tx, DerivationStrategyBase strategy, DerivationFeature feature)
 		{
 			return new Index(tx, $"{_Suffix}HighestPath", $"{strategy.GetHash()}-{feature}");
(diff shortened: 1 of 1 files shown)


NBXplorer/Repository.cs
```diff
@@ -1149,67 +1173,154 @@ public Task<TrackedTransaction[]> GetMatches(NBitcoin.Transaction tx, uint256 bl
 		{
 			return GetMatches(new[] { tx }, blockId, now, useCache);
 		}
+
+		private bool isTaproot(TxIn input)
+		{
+			// return true;
+			return input.ScriptSig.IsScriptType(ScriptType.Taproot);
```
NicolasDorier, line 1180: The whole problem is that this is not possible. You can't know if a txin is taproot just by looking the input.ScriptSig.

For this reason, I think taproot input/outputs should not be treated differently from the rest.

NBXplorer/Repository.cs
```diff
@@ -809,6 +809,30 @@ private static SavedTransaction ToSavedTransaction(Network network, ReadOnlyMemo
...
+		public async Task<MultiValueDictionary<OutPoint, KeyPathInformation>> GetKeyInformations(OutPoint[] outpoints)
+		{
+			// This is a unoptimized code, written so that it just gives the expected output.
+			// Should be replaced later on by a code like in GetKeyInformations(Script[] scripts)
+			MultiValueDictionary<OutPoint, KeyPathInformation> result = new MultiValueDictionary<OutPoint, KeyPathInformation>();
+			Dictionary<OutPoint, Script> outpointToScript = new Dictionary<OutPoint, Script>();
+			var scripts = new List<Script>();
+			foreach (var outpoint in outpoints)
+			{
+				var prevOutTxn = getTransaction(outpoint.Hash);
```
NicolasDorier, line 821: This is sadly not possible, the server would never been able to keep up with the blockchain if we needed 1 roundtrip for each outpoint.

NicolasDorier:
I think in the first PR you should completely disregard taproot, and just make sure to index both `scriptpubkey -> keyinfo` and `outpoint -> keyinfo`. Then trying to fetch both when looking for a match.

To see if it works correctly, you can comment out where you index `scriptpubkey -> keyinfo` for each Txout. Normally tests should still pass.

NBXplorer/Repository.cs
sageprogrammer: Okay will do that.

sageprogrammer: Okay will do that.

NBXplorer/Repository.cs
sageprogrammer: Initially I was also against the idea of making a round trip for each outpoint. But I could not find a function through which I could collectively query all the outpoints for the corresponding txns. This part of code was a sort of place holder for such a function.

NicolasDorier: not only this, but txindex is not mandatory for nbxplorer.

sageprogrammer: What do you mean by txindex not mandatory for nbxplorer?

sageprogrammer: Is it that we do not have to query for every txid we encounter, only the saved ones.

PR #329, opened 2021-08-14
Get KeyPathInformations from OutPoints

sageprogrammer:
Since in case of taproots it will not be possible to get `KeyPathInformation` from `SigScript` as was the case before. This pull request allows getting `KeyPathInformation`s from `OutPoint`s using 
- `OutPoint -> PubScript` : newly created index which stores a mapping of `OutPoint` and the corresponding `PubScript`.
- `PubScript -> KeyPathInformation`: already exists.

### DIFF
--- a/NBXplorer/Repository.cs
+++ b/NBXplorer/Repository.cs
@@ -666,10 +666,11 @@ namespace NBXplorer
 			await tx.Commit();
 		}
 
-		public async Task SaveOutPointToScript((OutPoint outPoint,Script script)[] mapping)
+		public async Task<MultiValueDictionary<OutPoint, KeyPathInformation>> SaveOutPointToScript((OutPoint outPoint, Script script)[] mapping)
 		{
 			var keyInfoByScripts = await GetKeyInformations(mapping.Select(m => m.script).ToArray());
 			var outPointWithKeyPath = new List<(OutPoint, KeyPathInformation)>();
+			var KeyPathByOutPoints = new MultiValueDictionary<OutPoint, KeyPathInformation>();
 			foreach (var map in mapping)
 			{
 				if (keyInfoByScripts.TryGetValue(map.script, out var keyPaths))
@@ -677,10 +678,12 @@ namespace NBXplorer
 					foreach (var keyPath in keyPaths)
 					{
 						outPointWithKeyPath.Add((map.outPoint, keyPath));
+						KeyPathByOutPoints.Add(map.outPoint, keyPath);
 					}
 				}
 			}
 			await SaveKeyInformations(outPointWithKeyPath.ToArray());
+			return KeyPathByOutPoints;
 		}
 
 		public async Task Track(IDestination address)
@@ -881,6 +884,16 @@ namespace NBXplorer
 					result.AddRange(outPoint, keyInfos);
 				}
 			}
+			return result;
+		}
+		public async Task<MultiValueDictionary<OutPoint, KeyPathInformation>> GetKeyInformations((OutPoint outPoint, Script script)[] outPointWithScript)
+		{
+
+			var result = await GetKeyInformations(outPointWithScript.Where(o => o.script is null).Select(o => o.outPoint).ToArray());
+			var leftOver = outPointWithScript.Where(o => !(o.script is null)).ToArray();
+			var leftOverResult = await SaveOutPointToScript(leftOver);
+			foreach (var kv in leftOverResult)
+				result.AddRange(kv.Key, kv.Value);
 			return result;
 		}
 		public async Task<MultiValueDictionary<Script, KeyPathInformation>> GetKeyInformations(Script[] scripts)
@@ -1298,15 +1311,13 @@ namespace NBXplorer
 					continue;
 				}
 				noMatchTransactions.Add(tx.GetHash());
-				var (txScripts, txTaprootOutpoints) = ExtractInfo(tx);
+				var txOutPointsWithScripts = ExtractInfo(tx);
 
-				foreach (var script in txScripts)
-					transactionsPerScript.Add(script, tx);
-				foreach (var input in tx.Inputs)
-					transactionsPerOutpoint.Add(input.PrevOut, tx);
-
-				scripts.AddRange<Script>(txScripts);
-				outPoints.AddRange(txTaprootOutpoints);
+				foreach (var outPointWithScript in txOutPointsWithScripts)
+				{
+					transactionsPerOutpoint.Add(outPointWithScript.outPoint, tx);
+				}
+				outPointsWithScripts.AddRange(txOutPointsWithScripts);
 			}
 
 			var keyPathInformationsByTrackedTransaction = new MultiValueDictionary<TrackedTransaction, KeyPathInformation>();
--- a/NBXplorer/Repository.cs
+++ b/NBXplorer/Repository.cs
@@ -1,4 +1,4 @@
-﻿using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Logging;
 using System.Linq;
 using NBitcoin;
 using System;
@@ -808,6 +808,30 @@ namespace NBXplorer
 			t.Timestamp = NBitcoin.Utils.UnixTimeToDateTime(timeStamped.TimeStamp);
 			t.Transaction.PrecomputeHash(true, false);
 			return t;
+		}
+		public async Task<MultiValueDictionary<OutPoint, KeyPathInformation>> GetKeyInformations(OutPoint[] outpoints)
+		{
+			// This is a unoptimized code, written so that it just gives the expected output.
+			// Should be replaced later on by a code like in GetKeyInformations(Script[] scripts)
+			MultiValueDictionary<OutPoint, KeyPathInformation> result = new MultiValueDictionary<OutPoint, KeyPathInformation>();
+			Dictionary<OutPoint, Script> outpointToScript = new Dictionary<OutPoint, Script>();
+			var scripts = new List<Script>();
+			foreach (var outpoint in outpoints)
+			{
+				var prevOutTxn = getTransaction(outpoint.Hash);
+				if (prevOutTxn != null)
+				{
+					scripts.Add(prevOutTxn.Outputs[outpoint.N].ScriptPubKey);
+					outpointToScript[outpoint] = prevOutTxn.Outputs[outpoint.N].ScriptPubKey;
+				}
+			}
+
+			var keyInformation = await GetKeyInformations(scripts.ToArray());
+			foreach (var outpoint in outpoints)
+			{
+				result.AddRange(outpoint, keyInformation[outpointToScript[outpoint]]);
+			}
+			return result;
 		}
 		public async Task<MultiValueDictionary<Script, KeyPathInformation>> GetKeyInformations(Script[] scripts)
 		{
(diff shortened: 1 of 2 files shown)


sageprogrammer:
I think we can rerun the tests, since I verified that build_and_test is passing in the local repo https://github.com/sageprogrammer/NBXplorer/pull/6

NBXplorer/Repository.cs
```diff
@@ -648,6 +652,40 @@ public async Task SaveKeyInformations(KeyPathInformation[] keyPathInformations)
...
+				return result;
+			foreach (var batch in outPoints.Batch(BatchSize))
+			{
+				using var tx = await engine.OpenTransaction();
+				foreach (var outPoint in batch)
+				{
+					var table = GetOutPointsIndex(tx, outPoint);
+					var rawScript = await table.SelectBytes(0);
+					if (rawScript is null)
+						result.Add(outPoint, null);
```
NicolasDorier, line 678: Given that the majority will end up in this case, you should not do this.
Imagine a block has around 3000 outpoints. This would add 3000 elements to the dictionary. (and with unset capacity on initialization, it will internally create lot's of copy of arrays, which will probably slow down things)

NBXplorer/Repository.cs
```diff
@@ -809,6 +856,26 @@ private static SavedTransaction ToSavedTransaction(Network network, ReadOnlyMemo
 			t.Transaction.PrecomputeHash(true, false);
 			return t;
 		}
+		public async Task<MultiValueDictionary<OutPoint, KeyPathInformation>> GetKeyInformations(OutPoint[] outPoints)
+		{
+			MultiValueDictionary<OutPoint, KeyPathInformation> result = new MultiValueDictionary<OutPoint, KeyPathInformation>();
+			if (outPoints.Length == 0)
+				return result;
+			var outPointToScript = await GetOutPointToScript(outPoints);
```
NicolasDorier, line 864: Since `GetOutPointToScript` is only used here, I would advise to get rid of it completely and iterates through the rows directly to this methods, adding to the `MultiValueDictionary` also the way.

NBXplorer/Repository.cs
```diff
@@ -1149,67 +1216,103 @@ public Task<TrackedTransaction[]> GetMatches(NBitcoin.Transaction tx, uint256 bl
...
-				{
-					var signer = input.GetSigner();
-					if (signer != null)
-					{
-						scripts.Add(signer.ScriptPubKey);
-						transactionsPerScript.Add(signer.ScriptPubKey, tx);
-					}
-				}
-				foreach (var output in tx.Outputs)
+				var txOutPointsWithScripts = ExtractInfo(tx);
```
NicolasDorier, line 1288: Since you use `ExtractInfo` only here, I would advise to just iterate through directly in this method, adding to `transactionsPerOutpoint` and `outPointsWithScripts` along the way.

NBXplorer/Repository.cs
```diff
@@ -1149,67 +1216,103 @@ public Task<TrackedTransaction[]> GetMatches(NBitcoin.Transaction tx, uint256 bl
...
+					continue;
+				outPointWithScript.Add((coin.Outpoint, coin.ScriptPubKey));
+			}
+			return outPointWithScript;
+		}
+
+		private void recordMatches(
+									IReadOnlyCollection<NBitcoin.Transaction> txs,
+									IReadOnlyCollection<KeyPathInformation> keyPathInformations,
+									ref MultiValueDictionary<TrackedTransaction, KeyPathInformation> keyPathInformationsByTrackedTransaction,
```
NicolasDorier, line 1242: `ref` are not needed in any of the args

NBXplorer/Repository.cs
```diff
@@ -1149,67 +1216,103 @@ public Task<TrackedTransaction[]> GetMatches(NBitcoin.Transaction tx, uint256 bl
...
-								new Dictionary<Script, KeyPath>());
-							match.FirstSeen = now;
-							match.Inserted = now;
-							matches.Add(matchesGroupingKey, match);
-						}
-						if (keyInfo.KeyPath != null)
-							match.KnownKeyPathMapping.TryAdd(keyInfo.ScriptPubKey, keyInfo.KeyPath);
-						keyPathInformationsByTrackedTransaction.Add(match, keyInfo);
-					}
+					recordMatches(
```
NicolasDorier, line 1304: `recordMatches` is not needed, we don't reuse it.

NicolasDorier:
I left several comments.


I see that you are saving `outpoint -> scriptPubKey`. This mean that when we scan a transaction, there is two query to the db: First getting the `scriptPubKey`, then getting the `KeyInformation`.

To prevent this, I thought de-normalizing by just having the mapping `outpoint -> KeyInformation`.

However, I think your solution is better, because in 99.9% of cases the second query will never happen as `outpoint -> scriptPubKey` will not match existing outpoint. I'll think about it a bit more, but seems ok.

NBXplorer/Repository.cs
sageprogrammer: Fixed this

NBXplorer/Repository.cs
sageprogrammer: Ok, I can remove the function. But it could be required in the future for other uses, right?

NBXplorer/Repository.cs
sageprogrammer: Fixed this

NBXplorer/Repository.cs
sageprogrammer: Removed the function altogether

NBXplorer/Repository.cs
sageprogrammer: Fidex this.

sageprogrammer:
> I left several comments.
> 
> I see that you are saving `outpoint -> scriptPubKey`. This mean that when we scan a transaction, there is two query to the db: First getting the `scriptPubKey`, then getting the `KeyInformation`.
> 
> To prevent this, I thought de-normalizing by just having the mapping `outpoint -> KeyInformation`.

Initially the plan was to make `outpoint -> KeyInformation` and due to which we will not have to do a join operation for getting the data. But later during the testing (For eg CanTrackAddress), It was found that after preparing `outpoint -> KeyInformation` using `scriptPubKey -> KeyInformation`, some times the same key-value pair in `scriptPubKey -> KeyInformation` got updated. Since we now had to update `outpoint -> scriptPubKey`, hence in each call we will have to check if it was updated or not, due to which it lost its advantage of being a single query. I think we might still be able to think of a better approach of updating the index regularly, but I went with `outpoint -> scriptPubKey` since even with join latency might not be that bad, and in terms of code it was better manageable.

sageprogrammer:
Here's an idea,
I am assuming that in DBtrie `scriptPubKey -> KeyInformation` when querying, it first finds the key and then returns the corresponding Value-reference stored against it. So if we update value, the reference itself does not change, but its internal value changes.

Then we could declare a function like `link(key1, key2)` where now both keys refer to the same value, 
So even if the value is updated through key `key1`, and now if we access through `key2`, it would still give us the updated value.
This could be used in case of scripts & outpoints.

Note: It will also add a little complexity when we are deleting a key-value in the DBTrie.

NicolasDorier:
Don't change what you did if it works. `outpoint -> scriptPubKey` latency problem might not be worth improving. Will poke around your PR

NBXplorer/Repository.cs
```diff
@@ -1149,67 +1218,73 @@ public Task<TrackedTransaction[]> GetMatches(NBitcoin.Transaction tx, uint256 bl
...
+					foreach (var input in tx.Inputs)
 					{
-						scripts.Add(signer.ScriptPubKey);
-						transactionsPerScript.Add(signer.ScriptPubKey, tx);
+						outPointsWithScripts.Add((input.PrevOut, null));
+						transactionsPerOutpoint.Add(input.PrevOut, tx);
 					}
 				}
-				foreach (var output in tx.Outputs)
+				foreach (var coin in tx.Outputs.AsCoins())
```
NicolasDorier, line 1246: Why are you iterating on coins? TxOut seems doing the same

NicolasDorier:
Improving a bit the code to limit allocations.

NicolasDorier:
I noticed that `public async Task<MultiValueDictionary<OutPoint, KeyPathInformation>> GetKeyInformations(IList<(OutPoint outPoint, Script script)> outPointWithScript)` is saving mapping.

This should not be the case, as it is called in GetMatches which is not meant to save anything. However, if I remove the `SaveOutPointToScript` then the code is breaking. This is not normal, trying to find out why...

sageprogrammer:
> I noticed that `public async Task<MultiValueDictionary<OutPoint, KeyPathInformation>> GetKeyInformations(IList<(OutPoint outPoint, Script script)> outPointWithScript)` is saving mapping.
> 
> This should not be the case, as it is called in GetMatches which is not meant to save anything. However, if I remove the `SaveOutPointToScript` then the code is breaking. This is not normal, trying to find out why...

There are only 2 places where the mapping is saved, `SaveTransactions` & `GetMatches`. 
In case of `GetMatches`, the above problem is caused because of the output side of the txns. So If we want to query using `OutPoint`, then we will have to save mapping. So if we do not want to save the mappings, we could directly query using `pubScriptKey`.

NicolasDorier:
@sageprogrammer taking a look at it, probably found a nice solution.

sageprogrammer:
Yup, Working on that.

NicolasDorier:
@sageprogrammer I did it, take a look at it, I think it works fine.

NicolasDorier:
Sorry, have forced push here.

The diff is so small.

NicolasDorier:
Ok I think that's good @sageprogrammer can you take a last look? I have removed as much diff as I could.

I think this solution of `oupoint -> scriptpubkey` is the right one, the code diff is really small. I am convinced now that improving roundtrip by having `outpoint -> keyinfo` would not be worse the complexity.

NBXplorer/Repository.cs
```diff
@@ -1181,10 +1223,20 @@ public async Task<TrackedTransaction[]> GetMatches(IList<NBitcoin.Transaction> t
...
+				{
+					transactionsPerScript.Add(kv.Value, tx);
+				}
+			}
 			if (scripts.Count == 0)
 				return Array.Empty<TrackedTransaction>();
 			var keyPathInformationsByTrackedTransaction = new MultiValueDictionary<TrackedTransaction, KeyPathInformation>();
-			var keyInformations = await GetKeyInformations(scripts.ToArray());
+			var keyInformations = await GetKeyInformations(scripts);
 			foreach (var keyInfoByScripts in keyInformations)
```
sageprogrammer, line 1240: I think we will also have to add the same for outpoints also

NicolasDorier: Look above, we are adding the scripts of the inputs in the list.

sageprogrammer: Just saw that, thats neat.

sageprogrammer: I also think this is good.

NicolasDorier:
Will merge this then!
Thanks for working on it, as I sent on mattermost:

I think the next step would be to migrate existing database.
The way to do this, is to detect whether the Outpoints table exists. And If not, iterating through all the `GetTransactionsIndex` and index all the `ReceivedCoins` to the table.
You can do the migration in the `RepositoryProvider.StartAsync`
NBX isn't available until it is done.

PR #330, opened 2021-08-20
Populate OutPoint table from existing Transactions table.

### LINKED ISSUE #325: Support for receiving on Taproot
Currently NBXplorer can doesn't allow the creation of a wallet on taproot. (or if it allows, it won't work correctly)

Sadly, to support it ,we need a refactoring of NBXplorer that will also require a migration step for existing installs, (though we can tackle in two different PRs)

The `GetMatches` methods, which scan a bunch of transactions for transactions that interest us is doing a hack which was supported to save a database query: From the scriptSig of a transaction input, we attempt to "guess" the scriptPubKey, then we query on that.

The problem is that it is impossible to do this with taproot, since the scriptSig does contain only a signature of 64 bytes.

Instead, what should be done, is to not only save the mapping `scriptPubKey -> KeyInformation` (as we do now), but also an additional mapping `outpoint -> KeyInformation`.

In the first PR, the GetMatches should still attempt to guess the scriptSig, even if it also query on the outpoint. The reason is that existing deployment would not work properly if we stopped doing this (we would miss some spending). The downside is increased pressure on the DB queries.

In a second PR, we can attempt to make a migration where we build the index `outpoint -> KeyInformation` from the information we already saved in DB.

sageprogrammer:
Closes #325 
This PR consists of
- Code to migrate data from existing Transactions table to newly created Outpoints table in #329 
- New tests for testing the migration
- New testing data for migration


NicolasDorier:
Awesome seems to be good to me.

I am a bit concerned for heavy users though, if their table is too big, then the update might not work using a single commit. (if it starts taking more than the DBCache, it would fail)

I think you can improve it by batching the update in several commit.
However, if it crashes on the middle, given you test the existence of the table we would end up in a half migrated database.

Let me think of a simple change that might solves those issues.

sageprogrammer:
> I think you can improve it by batching the update in several commit.

We could solve it by calling `tx.commit()` after processing a set number of txns, lets say `repo.BatchSize`.

> However, if it crashes on the middle, given you test the existence of the table we would end up in a half migrated database.

A naive solution can be, we can add a `checkpoint` in the `OutPoint` table, after saving a batch in the table.
The checkpoint could contain the number of txns processed. If the migration is done then we could store `"done"` in the `checkpoint`.
So when starting the migration, we could check the presence of `OutPoint` table. If it is present we check if it has a `checkpoint` field. If it contains `"done"`, then the migration was completed. If it is a number, then we know where to continue it from.

But this will not work if `Transactions` got changed after the crash.

NicolasDorier:
@sageprogrammer I don't think we need checkpoint, because the inserts are idempotent. If it fails, restarting from beginning is OK.

NicolasDorier:
I simplified just a bit. Seems good to go. I will try running that on some of my own production db.

NicolasDorier:
I found a bug, when the trackedtransaction doesn't have the full transaction (which happen for restored balances via scantxoutset)

Trying to add a test.

NicolasDorier:
@sageprogrammer I also fixed another bug: you were indexing all outputs of the transaction. This is overkill, we are only interested by the outputs we are tracking. This will be faster.

sageprogrammer:
Seems good to me.

PR #335, opened 2021-08-29
Add taproot derivation scheme

sageprogrammer:
This PR includes:

- Derivation scheme for Taproot.
- Modified the corresponding tests.
- Changed client target framework from `netstandard2.0` to `netstandard2.1`

### DIFF
--- a/NBXplorer.Tests/UnitTest1.cs
+++ b/NBXplorer.Tests/UnitTest1.cs
@@ -367,23 +367,24 @@ namespace NBXplorer.Tests
 				Assert.NotNull(spendingPSBT.Inputs[0].WitnessUtxo);
 				///////////////////////////
 
-				CanCreatePSBTCore(tester, true);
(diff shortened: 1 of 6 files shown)


NBXplorer.Tests/UnitTest1.cs
```diff
@@ -866,7 +867,7 @@ private static void CanCreatePSBTCore(ServerTester tester, bool segwit)
 			Assert.True(psbt2.PSBT.TryGetEstimatedFeeRate(out var feeRate));
 			Assert.Equal(new FeeRate(1.0m), feeRate);
 
-			if (segwit)
+			if (type == ScriptPubKeyType.Segwit)
```
NicolasDorier, line 870: this should also pass with taproot.

NicolasDorier:
I will make a commit to keep supporting netstandard2.0, but disabling taproot related stuff if that's the case.

PR #356, opened 2022-02-28
Implement a postgres backend

NicolasDorier:
NBXPlorer is currently depending on DBTrie, which is an embedded no sql database.
The lack of flexibility regarding queries means users start hitting performance issues when reaching around 30 000 UTXO for a wallet. Even if it still work even up to 100 000 UTXO, the experience is degraded.

This is working in progress implementing a postgres backend aimed at allowing wallets even with 200 000 UTXO to perform smoothly.
On top of this, we expect users to take advantage of the full power of SQL to make their own queries rather than going through NBXPlorer.

This new backend will be supported by the current API of NBXplorer. But a new api will be built to take full advantage of the more flexible queries NBXplorer can make.

We also plan to implement a feature to migrate the data from the old backend to postgres.
The old backend will still be supported for a few years, but don't expect new feature there.

This new backend also mean we won't be storing the block headers  of the chain in memory, thus decreasing significantly the RAM required by NBXPlorer.


marceloceccon:
This is awesome. 
I had just created a fork to add a method to query if an address is tracked but with the postgres feature I can query directly from the database. Will keep an eye on this branch and test it as well.

NicolasDorier:
So the tests are going well:

The balance and UTXO calculation is propertional to the number of utxos rather than the history of the wallet! For a wallet with 200K UTXOs, it takes 2sec to calculate the balance. For 1 UTXO, it is immediate!

marceloceccon:
If you ping me when you think the fork is ready for testing, I can run this version on the staging environment of an exchange. I currently have nbxplorer watching btc, ltc & doge and monitor the server so I will be able to check some metrics on cpu usage/io profile of the machine with both versions.

NicolasDorier:
@marceloceccon thank you, I will definitely need feedback.
I still need to do the migration code though. Should be done this week.

NicolasDorier:
@[marceloceccon](https://github.com/marceloceccon) can you test it?
The schema is not fixed yet, and I will probably change a few things still, but I'd like some feedback.

You need to start with `NBXPLORER_POSTGRES="connectionString"`, and `NBXPLORER_AUTOMIGRATE="true"`.

You can try a few views like `wallets_balances` and `wallets_utxos`, also any other routes of NBX API should be now way faster.
@farukterzioglu  can you give it a try as well?

marceloceccon:
This is great news. I will try over the weekend. The memory usage will be so much more under control (because one can of course limit postgres's memory) and with a fast storage and indexes the lookups will be fast. Because I don't track by derivation, I track specific addresses and to get a dense utxo database fill I am getting 1.000.000 addresses that had transactions in the last few blocks.

I wrote a file with 1.000.000 btc addresses that have transacted within the recent blocks, you can get it here if you want to use for testing as well, I am going to benchmark using these addresses.

 https://github.com/marceloceccon/epheremalpaste/blob/master/addresses_with_transactions_recent_blocks_march_2022.txt

Also, check this out, pretty interesting graph on UTXO age.  https://unchained.com/hodlwaves/

farukterzioglu:
I encountered a problematic case, but not sure about the cause. 
I tried two more times to be sure. Here is the case;

I started a clean NBXplorer setup from this branch. (without Postgres )
I created a new wallet, get a new address, sent 2 txs, created a block, sent 1 more tx.
When I query them I got 2 confirmed txs, 1 unconfirmed tx, as expected. 

I started it again ONLY with `NBXPLORER_POSTGRES` value
It created a database and the tables;
image

I stopped it and started it again with both `NBXPLORER_POSTGRES` and `NBXPLORER_AUTOMIGRATE` values. 
But this time it throws an error as;
image

But if I try it with both `NBXPLORER_POSTGRES` and `NBXPLORER_AUTOMIGRATE` values at the first start, it migrates it successfully.

farukterzioglu:
Got an idea about the above, when I start it without `NBXPLORER_AUTOMIGRATE` value, it still creates a record in `blks` table;

image

When I restart it again with `NBXPLORER_AUTOMIGRATE` I guess it tries to insert it again.

farukterzioglu:
Another issue. 
For the migration of production data, I want to test it without node connection. 

I will get the production data to my local environment, remove node connection settings (because my local env doesn't have access to live nodes), I will trigger the migration, then check it with queries. 

But when I try it without node connection settings, migration fails; 
image

farukterzioglu:
It will be take more time to test it but is it ready to be tested for '1 NBX - multiple blockchains' setup 
e.g. `--chains=btc,ltc,dash`

NicolasDorier:
@farukterzioglu I already fixed this issue about the block not found, just pull

NicolasDorier:
> I started a clean NBXplorer setup from this branch. (without Postgres )
> I created a new wallet, get a new address, sent 2 txs, created a block, sent 1 more tx.
> When I query them I got 2 confirmed txs, 1 unconfirmed tx, as expected.

This is expected. As the migration need to be done on a fresh database. I can try to improve the error message though.

NicolasDorier:
> But when I try it without node connection settings, migration fails;

It's really better that you have a full node here. But I can try to find a work around.

NicolasDorier:
@farukterzioglu can you try again without RPC? I made the import fallback on the `[cryptocode]slim-chain.dat` if RPC isn't available.

NicolasDorier:
@marceloceccon about the memory usage, while it will be better since there is less allocation than before, we are still loading the block headers in memory.

I will try to remove this later if postgres is used. I feel it is possible, but need a bit more work.

farukterzioglu:
I am able to migrate it without rpc connections, thanks. 

it may be expected but I saw something; 
When I migrate without rpc connection, previously confirmed transactions became unconfirmed, but if I restart it with rpc connection, they become confirmed. btw I didn't generate new block.

NicolasDorier:
@farukterzioglu that's weird, can you try to repro it?
Txs should be unconfirmed only if the blocks failed to be imported in the db. (as such, blks and blks_txs table should be empty)

farukterzioglu:
@NicolasDorier  I tried again and the result is same but it looks like it is not related to the migration, I explained at the end.

These are the steps and the result;

I started with latest commit from `postgres` branch without Postgres. 
I sent 3 txs, generated a block, sent 1 txs. When I query I can see 3 conf., 1 unconf. transactions. 

Then I migrated them **without** rpc connection. Tables and records are as expected;
txs
image

blks
image

blks_txs
image

But when I query them with `/v1/cryptos/btc/derivations/tpub....-[p2sh]/transactions` ;

image
All 4 txs are unconfirmed. 

I deleted the migration_lock file, started again without Postgres to check (with rpc);
image
Same as before the migration; 3 conf. txs, 1 unconf. txs.

Then again, I migrated them **without** rpc, result same as before: all unconfirmed. 
Then I restarted it with Postgres and **with** rpc, response is correct,3 conf./1uncofirmed. 
Then I restarted it **without** rpc, and all txs are unconfirmed.

Looks like it is not related to the migration or to the data on the db, since they are correct. 


I tried the same scenario, starting with/without rpc, with the legacy db and the result is same.
With rpc I am getting 3 conf./1unconf. txs, without rpc they became all unconfirmed. 

Looks like it is an existing behavior not related to Postgres.

NicolasDorier:
@farukterzioglu I see. The thing is that the /transactions still depends on the `slim-chain.dat` and `Index Progress` to advertise confirmation, rather than the database. This is something I want to eventually change.

The code path for /transactions is the same as before, compared with /utxos and /balances which just query the db.

Normally this should eventually sort out when the migrated node catch up the sync.

NicolasDorier:
@farukterzioglu I tried to repro in a test, didn't manage.

What the test do: 
* Alice get 3 utxos: one confirmed in b1, one confirmed in b2, and an unconfirmed
* b2 becomes invalid block
* nbx restart with automigrate and postgres
* both, the /transactions and the /utxos report properly only a single utxos/transaction from b1.

NicolasDorier:
Cool. Let me know how that goes about perf on a production like load.

NicolasDorier:
I tested on the DB of https://mainnet.demo.btcpayserver.org/ 
The size of the DBTrie DB is 5GB, the migration took 53min. The postgres DB is 1.7GB, requests are instant. RAM usage is 66MB.

NicolasDorier:
@farukterzioglu @junderw @marceloceccon . This is a good time to test this PR by migrating.

I added doc on https://github.com/dgarage/NBXplorer/blob/454088c5577663ed4d2e1e6c918a48f912812668/docs/Postgres-Migration.md

Please let me know how it goes, I worked a lot to try to make it as fast as possible.

NicolasDorier:
I documented the useful views on https://github.com/dgarage/NBXplorer/blob/6eb9214f42d8dd0f895e2818edbc023a8c9955e7/docs/Postgres-Schema.md

farukterzioglu:
I tried 3 times to migrate our 15GB data but all failed because of a timeout exception. 
Last time I tried with ` --nomigrateevts --nomigraterawtxs` parameters.

Exception is 
```
Unhandled exception. Npgsql.NpgsqlException (0x80004005): Exception while reading from stream
 ---> System.TimeoutException: Timeout during reading attempt
   at Npgsql.Internal.NpgsqlConnector.<ReadMessage>g__ReadMessageLong|211_0(NpgsqlConnector connector, Boolean async, DataRowLoadingMode dataRowLoadingMode, Boolean readingNotifications, Boolean isReadingPrependedMessage)
   at Npgsql.NpgsqlDataReader.NextResult(Boolean async, Boolean isConsuming, CancellationToken cancellationToken)
   at Npgsql.NpgsqlCommand.ExecuteReader(CommandBehavior behavior, Boolean async, CancellationToken cancellationToken)
   at Npgsql.NpgsqlCommand.ExecuteReader(CommandBehavior behavior, Boolean async, CancellationToken cancellationToken)
   at Npgsql.NpgsqlCommand.ExecuteNonQuery(Boolean async, CancellationToken cancellationToken)
   at Dapper.SqlMapper.ExecuteImplAsync(IDbConnection cnn, CommandDefinition command, Object param) in /_/Dapper/SqlMapper.Async.cs:line 654
   at NBXplorer.HostedServices.DBTrieToPostgresMigratorHostedService.Migrate(NBXplorerNetwork network, Repository legacyRepo, PostgresRepository postgresRepo, ILogger logger, CancellationToken cancellationToken) in /NBXplorer/NBXplorer/HostedServices/DBTrieToPostgresMigratorHostedService.cs:line 463
   at NBXplorer.HostedServices.DBTrieToPostgresMigratorHostedService.StartAsync(CancellationToken cancellationToken) in /NBXplorer/NBXplorer/HostedServices/DBTrieToPostgresMigratorHostedService.cs:line 137
   at Microsoft.AspNetCore.Hosting.HostedServiceExecutor.ExecuteAsync(Func`2 callback, Boolean throwOnFirstFailure)
   at Microsoft.AspNetCore.Hosting.WebHost.StartAsync(CancellationToken cancellationToken)
   at Microsoft.AspNetCore.Hosting.WebHostExtensions.RunAsync(IWebHost host, CancellationToken token, String startupMessage)
   at Microsoft.AspNetCore.Hosting.WebHostExtensions.RunAsync(IWebHost host, CancellationToken token, String startupMessage)
   at Microsoft.AspNetCore.Hosting.WebHostExtensions.RunAsync(IWebHost host, CancellationToken token)
   at Microsoft.AspNetCore.Hosting.WebHostExtensions.Run(IWebHost host)
   at NBXplorer.Program.Main(String[] args) in /NBXplorer/NBXplorer/Program.cs:line 60
```

And progress is as;
image

However, when I try to run it without migration, it runs without problem, but I am not sure if there is anything missing.
image

NicolasDorier:
Reminder to myself: settled issue with farukterzioglu on the mattermost chat.

farukterzioglu:
Should we also add note for whom will encounter at the future about timeouts at DASH, that we fixed just by `ANALYZE` postgres command? 
Or it won't happen again for anyone?

PR #377, opened 2022-09-29
Add InputIndex to matched inputs

Kukks:
Alternative to #370 where we add a dedicated prop instead of replacing the output index prop


NicolasDorier:
`IndexOfInput` would throw an exception if the transaction has been pruned (would happen if you restore a wallet)

NicolasDorier:
I think you should fix that in `PostgresRepository.GetTransactions` I believe that `utxo.idx` has the value you want.
The fix wouldn't work for DBTrie but we don't care.

PR #410, opened 2023-03-20
Marceloceccon patch lower cpu usage index

marceloceccon:
@NicolasDorier I believe the [015.CreateIndexFasterMatches.sql] may be interesting for everyone. CPU usage savings on the database server was dramatic. The CONCURRENTLY also makes it not lock the table whilst creating so it can be run on an live environment.

The other commits where I have the /istracked route probably not.

### DIFF
--- a/NBXplorer.Client/Models/IsTrackedResponse.cs
+++ b/NBXplorer.Client/Models/IsTrackedResponse.cs
@@ -1,0 +1,10 @@
+using NBXplorer.Models;
+
+namespace NBXplorer.Models
+{
+	public class IsTrackedResponse
+	{
+		public bool IsTracked { get; set; }
+		public TrackedSource TrackedSource { get; set; }
+	}
+}
--- a/NBXplorer.Tests/AnalysisTests.cs
+++ b/NBXplorer.Tests/AnalysisTests.cs
@@ -135,7 +135,7 @@ namespace NBXplorer.Tests
 				catch { }
 			}
 			using var client = new HttpClient();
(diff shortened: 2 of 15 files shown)


NicolasDorier:
Didn't see your PR sorry for the delay (please ping me on twitter DM if I miss it too long)

1. Queries shouldn't assume the schema of metadata. The whole point of this column is to have arbitrary values.
2. Why are you doing this query? `SELECT 1 FROM wallets_scripts WHERE code=@code AND script=@script AND wallet_id=@wid` works and is already indexed.

marceloceccon:
No worries. So I actually should have made two different PRs. One for the /istracked functionality and the other for the index.

So the index based on a fresh install from the root repository. So after I install and register a couple of thousand addresses to be tracked, the CPU usage would be very high. I then used pghero to monitor the offending queries and experimented with a couple of indexes until I got it and cpu usage is minimal after the index.

> 1. Queries shouldn't assume the schema of metadata.

I am sorry I don't understand where does that apply.

NicolasDorier:
You didn't reply why you don't use this query to achieve same result: `SELECT 1 FROM wallets_scripts WHERE code=@code AND script=@script AND wallet_id=@wid`

marceloceccon:
I am sorry for the delay. It turns out this index was most likely related to my specific fork that had a /istracked api route which allowed the system to check if an address was already tracked and not applicable to the main and this is called by our systems frequently. 

I am closing this PR.

Issue #411, opened 2023-03-23
Add `replacing` property to events


kallerosenbaum:
I suggest adding `replacing` property to events. This would make users' lives easier and safer.

There are two issues with the `replacing` and `replacedBy` properties:

1. They aren't included (both are null) when you query specific transactions, e.g. `HTTP GET v1/cryptos/{cryptoCode}/derivations/{derivationScheme}/transactions/{txId}`. If you have lots and lots of transactions for your derivation scheme, querying all transactions when you're only interested in one becomes really impractical.
2. They aren't included in `newtransaction` events (I guess `replacedBy` isn't applicable, but `repacing` certainly is)

For me, fixing issue 2 would be preferable, but if that's not possible, fixing 1 works too. Of course both would be great.

NicolasDorier:
will take a look.

NicolasDorier:
@kallerosenbaum so sadly, the way `v1/cryptos/{cryptoCode}/derivations/{derivationScheme}/transactions/` works is by querying all the transactions in memory and finding what is relpacing what. So even if you had one transaction, if you want 1, I'd need to take all from DB. This only concerned the `replacing` property.

I may be able to hack something for `replacedBy`, but I'd prefer not, this code is a bit messy. (only if you think you REALLY need it)

That said, I am adding it to `cryptos/{cryptoCode}/transactions/{txId}`, so that might be enough for you. (only `replacedBy`)

Btw, did you consider querying the postgres tables directly?
The schema is straightforward, so in some case it might be more performant and flexible for you.

NicolasDorier:
Trying to add `replacing` to the event as well...

NicolasDorier:
@kallerosenbaum I have another trouble: `replacing` should be an array rather than a single value. As a single transaction can replace sevaral other transaction.

NicolasDorier:
Please review the two PRs documentation if that's all you need.

kallerosenbaum:
I have reviewed now, only one question, on #413 . Thanks so much for your efforts!!!

PR #413, opened 2023-03-24
Add replaced tx ids in transaction events

NicolasDorier:
Address https://github.com/dgarage/NBXplorer/issues/411

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -696,7 +696,8 @@ Then you will receive such notifications when a transaction is impacting the `de
         "value": 100000000
       }
     ],
-    "cryptoCode": "BTC"
+    "cryptoCode": "BTC",
+    "replacing": ["25d6bc1b2812670550aca8b2984670203b5ebf00e75f9b2bbf1940c3fa27841e", "81a20eb55ec16b92c65d4e142278fd521caa9e5dcad9d941c8e256dbd917ae84"]
   }
 }
 ```
@@ -724,6 +725,10 @@ If you want all transactions of all derivation schemes of all crypto currencies,
 ```
 
 As an alternative to get notification, you can also use long polling with the [event stream](#eventStream).
+
+Fields:
+
+* `replacing`: The list of the unconfirmed transactions of this wallet which have been replaced by this new transaction. This can typically be used to detect when the sender is bumping fee. This can't be used to detect when the sender is attempting to abort a transaction.
 
 ## <a name="broadcast"></a>Broadcast a transaction
 
--- a/NBXplorer.Client/Models/NewTransactionEvent.cs
+++ b/NBXplorer.Client/Models/NewTransactionEvent.cs
@@ -37,6 +37,8 @@ namespace NBXplorer.Models
 		[JsonIgnore]
 		public override string EventType => "newtransaction";
 
+		public List<uint256> Replacing { get; set; }
+
 		public override string ToString()
 		{
 			var conf = (BlockId == null ? "unconfirmed" : "confirmed");
--- a/NBXplorer.Tests/Extensions.cs
+++ b/NBXplorer.Tests/Extensions.cs
@@ -13,15 +13,15 @@ namespace NBXplorer.Tests
 {
 	public static class Extensions
 	{
-		public static void WaitForTransaction(this LongPollingNotificationSession session, DerivationStrategyBase derivationStrategy, uint256 txId)
+		public static NewTransactionEvent WaitForTransaction(this LongPollingNotificationSession session, DerivationStrategyBase derivationStrategy, uint256 txId)
 		{
-			session.WaitForTransaction(TrackedSource.Create(derivationStrategy), txId);
+			return session.WaitForTransaction(TrackedSource.Create(derivationStrategy), txId);
 		}
-		public static void WaitForTransaction(this LongPollingNotificationSession session, BitcoinAddress address, uint256 txId)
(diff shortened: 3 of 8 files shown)


docs/API.md
```diff
@@ -696,7 +696,8 @@ Then you will receive such notifications when a transaction is impacting the `de
         "value": 100000000
       }
     ],
-    "cryptoCode": "BTC"
+    "cryptoCode": "BTC",
+    "replacing": ["25d6bc1b2812670550aca8b2984670203b5ebf00e75f9b2bbf1940c3fa27841e", "81a20eb55ec16b92c65d4e142278fd521caa9e5dcad9d941c8e256dbd917ae84"]
```
kallerosenbaum, line 700: Is this only for events over websocket or does it apply to long polling as well? If yes, then this is *exactly* what I need. I use long polling.

NicolasDorier: yes it does

NicolasDorier:
@kallerosenbaum is it really what you need? because newtransaction only fire for transactions impacting a wallet tracked by NBX.

Say Alice send to Bob, and only Bob is tracked. If Alice RBF to send to Carole instead, you won't see any `newtransaction`.

kallerosenbaum:
@NicolasDorier yes, I do think that's what I need. I have a polling mechanism that checks for tx presence on the node (that has `txindex=1`) via `v1/cryptos/{cryptoCode}/rpc` endpoint. When the poller finds out that a tx is gone (with a few seconds delay due to polling interval), I know that particular tx is gone, but I don't know whether its been replaced by another tx to me or if I'm being scammed, or because some other issue.

If I get the `replacing` property on events, I can immediately turn off polling for the replaced transaction and be sure that the transaction is replaced by this new transaction, for which I will start polling for presence.

Also, knowing that I get the `replacing` property on `newtransaction` events, if the poller detects a tx is gone, I know for sure that that tx isn't being replaced with another tx that pays me, so I know that there's no money for me in the mempool.

> instead a separate event might be needed
Hmm, you mean like a `gonetransaction` event that specifies whether it's been replaced by another tx that sends funds to us, or not.

NicolasDorier:
@kallerosenbaum yes. I fear that with the current API of this change, people will start to rely on it for detecting double spending, when there are double spending case that would be unnoticed.

kallerosenbaum:
@NicolasDorier if you are able to implement a `gonetransaction` event that'd be absolutely fantastic. That'd mean I can kill quite a bit of code from my client. Here are my 2sats on how I'd like to see it implemented.

Definitions:
"Honest Replacement": a transaction is replaced by another transaction that pays to the same address (not necessarily same amount)
"Double Spend": any other replacement

I think a `gonetransaction` event should be used only for Double Spends. And that a Honest Replacement should be communicated through a `replacing` property of `newtransaction` as in this PR.

I suggest this because from my POV, it'd be really helpful if an event communicates this information atomically, i.e. I get to know whether it's an Honest Replacement or a Double Spend in one single event.

NicolasDorier:
@kallerosenbaum feel free to ping me on twitter DM.

There is a way I detect reliably the double spending, but that isn't "real time".

I am automatically rebroadcasting transactions, and I can see from the result of Bitcoin Core if it has been double spent.
However, I can't know the transaction id that double spent it.

Would that be enough?

docs/API.md
Lightheriver: Hi i have no idea of what im doing or what your talking about  if u can only help me ou t ill really appreciate it i dont know if this is real or is fraud

Lightheriver (commented):
Hi

kallerosenbaum:
@NicolasDorier sorry for late answer. I kind of lost track of what your current proposal is, so let me rephrase my understanding of it, given my definitions of "double spend" and "honest replacement":

* When a honest replacement takes place, a newtransaction is issued as is already the case today, but with a new `replacing` property that reveals what other transactions it replaces.
* When a double spend occurs, a `gonetransaction` is issued, saying what txid has gone missing.

Is this correct? If so, yes, it's perfect.

NicolasDorier:
@kallerosenbaum yes, however the `gonetransaction` wouldn't be real time. It only happen when we try and fail to rebroadcast.

kallerosenbaum:
> @kallerosenbaum yes, however the `gonetransaction` wouldn't be real time. It only happen when we try and fail to rebroadcast.

@NicolasDorier yes. That's totally acceptable, especially if the retry interval is configurable.

docs/API.md
```diff
@@ -725,6 +726,10 @@ If you want all transactions of all derivation schemes of all crypto currencies,
 
 As an alternative to get notification, you can also use long polling with the [event stream](#eventStream).
 
+Fields:
+
+* `replacing`: An array of all transaction ids which have been directly or indirectly replaced by this transaction.
```
kallerosenbaum, line 731: Maybe a note here, saying something like: "This can be used to detect when an output for a tracked address is replaced by a transaction with an output *for that same address*, regardless of amount."

I'm not sure if "for that same address" is true, maybe it should say "for any tracked address" or "for any address in the same derivation scheme", but it should specify the conditions for this property being set.

kallerosenbaum:
Tested eb5847b19a554616df72f448c240bec1f1af5c45. Works as expected.

docs/API.md
NicolasDorier: Reworded:

`The list of the unconfirmed transactions of this wallet which have been replaced by this new transaction. This can typically be used to detect when the sender is bumping fee. This can't be used to detect when the sender is attempting to abort a transaction.`

Issue #422, opened 2023-08-29
substractFees removes the destination from outputs if amount is too low


farukterzioglu:
When we try to create a PSBT with a small amount (like 700) with the option `substractFees`, because the resulting amount (700 - fee) will be less than dust, it removes this output and transaction ends up with only change output. 

E.g.: 

Request;
```
{
    "Destinations": [
        {
            "Destination": "2N4mzNV2....",
            "Amount": 700,
            "SubstractFees": true
        }
    ]
}
```
Response; 
```
{
    "spentCoins": [
        {
            "keyPath": "0/0",
            "value": 612,
            "outpoint": {
                "hash": ".....b202f58b3451c8b4416....",
                "index": 1
            }
        },
        {
            "keyPath": "0/0",
            "value": 564,
            "outpoint": {
                "hash": "....a4ba9004117a7478e195a....",
                "index": 1
            }
        }
    ],
    "changeInformation": {
        "keyPath": "1/17",
        "value": 951
    },
    "transaction": "0100000002088826b4....08700000000",
    "fee": 225,
    "vSize": 225
}
```

Decoded tx;
```
{
    "addresses": [
        "2NAdn4bW53....",
        "2NDyGnkbTY...."
    ],
    "block_height": -1,
    "block_index": -1,
    "confirmations": 0,
    "double_spend": false,
    "fees": 225,
    "hash": "624eaaa823a13b2cd20cc906c7570cfd1f11dfc5db324f47a50175f9648229b4",
    "inputs": [
        {
            "addresses": [
                "2NAdn4bW53..."
            ],
            "age": 2475275,
            "output_index": 1,
            "output_value": 612,
            "prev_hash": "a933a3c4508e4add3....",
            "script": "16001481e8ce7654b0c....",
            "script_type": "pay-to-script-hash",
            "sequence": 4294967295
        },
        {
            "addresses": [
                "2NAdn4bW53...."
            ],
            "age": 2475280,
            "output_index": 1,
            "output_value": 564,
            "prev_hash": "b7ca36c949fc0b206bc....",
            "script": "16001481e8ce7654....",
            "script_type": "pay-to-script-hash",
            "sequence": 4294967295
        }
    ],
    "outputs": [
        {
            "addresses": [
                "2NDyGnkbTYn9Kn...."
            ],
            "script": "a914e356530a7b378.....",
            "script_type": "pay-to-script-hash",
            "value": 951
        }
    ],
    "preference": "low",
[...]

PR #425, opened 2023-09-12
Output below dust

farukterzioglu:
Fixes https://github.com/dgarage/NBXplorer/issues/422

### DIFF
--- a/NBXplorer.Tests/UnitTest1.cs
+++ b/NBXplorer.Tests/UnitTest1.cs
@@ -448,7 +448,7 @@ namespace NBXplorer.Tests
 					catch (NBXplorerException ex)
 					{
 						if (i == 25)
-							Assert.Equal("output-too-small", ex.Error.Code);
+							Assert.Equal("not-enough-funds", ex.Error.Code);
 						else
 							throw;
 					}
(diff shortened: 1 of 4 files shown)


NBXplorer.Tests/UnitTest1.cs
```diff
@@ -777,7 +777,8 @@ private static void CanCreatePSBTCore(ServerTester tester, ScriptPubKeyType type
 			Assert.Equal(outpoints[0], actualOutpoints[0]);
 			request.MinValue = Money.Coins(0.1m);
 			ex = Assert.Throws<NBXplorerException>(() => tester.Client.CreatePSBT(userDerivationScheme, request));
-			Assert.Equal("not-enough-funds", ex.Error.Code);
+			Assert.Equal("output-too-small", ex.Error.Code);
```
NicolasDorier, line 780: this seems wrong to me.
`MinValue` is restricting the selected UTXOs. So the error message should be `not-enough-funds`, not `output-too-small`.

NicolasDorier: So I looked over this case and while this is a breaking change, this isn't really illogical.

The PSBT is a "SweepAll", but there isn't any coin left in the wallet after the `MinCoins` set above.
So "Send all to address X" is basically "Send 0 to address X" which is indeed an "output that is too small".

NicolasDorier: I will fix this PR to keep compatibility with the previous behavior, as it makes sense to say that there isn't enough funds to sweep all.

PR #430, opened 2023-10-03
Fix big endian problem for numerics

### LINKED ISSUE #428: Incorrect balance when NBXplorer is running on Linux s390x Server
We are running NBXplorer and Postgres database on Linux s390x server. 
Balance data is correct in DB, and also when we get transactions with `/transactions` every thing is correct. But when we get the balance with `/balance`, it returns a very big number, hexadecimal presentation of actuall balance value.

If we run the same NBX in another place (non s390x) it works ok. 

Here are some data;

from `nbxv1_tracked_txs`;
`BTC  | e86117a46a211b2ccacadf22a1689d07a4946b4eb2c3e8dc6be221e4ae07ad15 |   1 | 1030678 | 2MvA9XerLqnRKj1UubEu8D5pnqJrNNtk8Nk`

`/transactions` response;

```
{
    "confirmedTransactions": {
        "transactions": [
            {
                "transactionId": "e86117a46a211b2ccacadf22a1689d07a4946b4eb2c3e8dc6be221e4ae07ad15",
                "outputs": [
                    {
                        "keyPath": "0/0",
                        "scriptPubKey": "a9141ff29342852c62943f72caac56bb22f5f08bb3a487",
                        "index": 1,
                        "value": 1030678,
                        "address": "2MvA9XerLqnRKj1UubEu8D5pnqJrNNtk8Nk"
                    }
                ],
                ...
            }
        ]
    }
    ...
}
```

balance request;  
`HTTP GET /v1/cryptos/BTC/derivations/tpubD6NzVbkrYhZ4XfMuYMcbbY2MS7gMksCntMajRsimXp9Zk7CPd1xNXRsMkqeiZBReuGNGPnspk6Tc7k2cJ3kHM8o9KTHdffhMHaEys4iKwtR-%5Bp2sh%5D/balance`
response;
```
{"unconfirmed":0,"confirmed":4426728302706688,"total":4426728302706688,"immature":0,"available":4426728302706688}
```

watch out for `4426728302706688` in balance. 

decimal value: `1030678`, hexadecimal equivalent: `FBA16`
also;
`4426728302706688` is **FBA16**00000000 in hexadecimal

I added some logs to understand when exactly it get wrong, it seems as soon as we get it from DB;
image

image

farukterzioglu:
Fixes #428

### DIFF
--- a/NBXplorer/Controllers/PostgresMainController.cs
+++ b/NBXplorer/Controllers/PostgresMainController.cs
@@ -48,7 +48,13 @@ namespace NBXplorer.Controllers
 			var network = GetNetwork(cryptoCode, false);
 			var repo = (PostgresRepository)RepositoryProvider.GetRepository(cryptoCode);
 			await using var conn = await ConnectionFactory.CreateConnection();
-			var b = await conn.QueryAsync("SELECT * FROM wallets_balances WHERE code=@code AND wallet_id=@walletId", new { code = network.CryptoCode, walletId = repo.GetWalletKey(trackedSource).wid });
+			var b = await conn.QueryAsync(@"SELECT 
+				wallet_id, code, asset_id,
+				CAST(confirmed_balance AS VARCHAR), 
+				CAST(unconfirmed_balance AS VARCHAR),  
+				CAST(available_balance AS VARCHAR),
+				CAST(immature_balance AS VARCHAR)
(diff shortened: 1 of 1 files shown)


farukterzioglu:
Since this is not a problem of NBX but Npgsql's, I will cancel this MR. 
In case someone need to run NBX on a S390X platform until the issue gets fixed, this workaround can be applied on a private fork.

Issue #291, opened 2020-08-21
Feature Request: Address groups


junderw:
This feature will be implemented in the API as two parts:

1. A request parameter when tracking an individual address.
2. A new derivationStrategy type for all HD based calls.

## Tracking

When tracking an individual address, passing an optional `group` value (string) will set a grouping for the address. (ie. "xyz").

So if I call track 5 times with 5 different addresses but all 5 times I add ?group=xyz, then all 5 of those addresses will be contained within group xyz.

## HD based queries

In addition to current derivationStrategy, a new type will be created:

`group-[xyz]` where xyz is the name of the group.

Any time a call is made using `group-[xyz]` as the dS, it will perform the query as-if the addresses that were added into the group all came from the same derivationStrategy.

sageprogrammer:
@NicolasDorier Is this still a required feature

NicolasDorier:
I think it is useful feature yes

sageprogrammer:
Has there been work done on this? If not I would like to work on it.

junderw:
I don't think anyone has worked on this.

Kukks:
I will begin work on something along these lines.

Kukks:
> I will begin work on something along these lines.

Update on this:
So it turns out this is quite easy to do and supported on the postgres backend already

junderw:
This seems like it could just be added as some extra conditions on the track endpoint.

PR #431, opened 2023-10-05
NBX enhancements

### LINKED ISSUE #429: Import UTXOs endpoint
In the case that you are starting to track an existing wallet, where you have its current state, we should be able to import it into NBX.

Given a tracked source, I have a set of coins it has received and its last index(es) used to derive addresses, I should be able to tell NBX to import these utxos and to set the index counter to prevent address reuse. 

NBX is able to verify these utxos are valid in a lightweight manner by calling gettxout on bitcoin core.

Kukks:
closes #429  #291 

This PR makes use of the enhanced flexivility of the postgres indexer and exposes them over the current API.

Notably it:
* has a new tracked source, Wallet, which can act as a "bag" for individual scripts, or as a parent to other wallets.
* the generate wallet and track wallet now support specifying a parent wallet
* there is an "associate scripts" endpoint which adds specific scripts/addresses to a tracked source to be watched.
* there is an "import coins" endpoint which allows you to tell NBX that a wallet has existing coins that it has not managed to track (as they were possibly created before nbx started tracking the wallet). This endpoint verifies the coins are real by calling gettxout on bitcoin core. If you have a local state of coins, this is a much lighter and scalable version of resotring a wallet's utxos on nbx than the scanutxoset option (which can only run exclusively and takes a lot of resources). This can be improved by also adding a utxoproof that it was mined in a specific block.


NicolasDorier:
tests seems to fail

NicolasDorier:
supersede by https://github.com/dgarage/NBXplorer/pull/450

There isn't any change, it's just a rebase.

PR #439, opened 2023-11-14
Remove DBTrie

NicolasDorier:
Completely remove the DBTrie backend from the source code.


bitcoinbrisbane:
Epic!  Good work.  Wonder if its worth a BIP on the SQL schemea?

NicolasDorier:
I think I will kick the can for this one.
The new .NET 8.0 required a few refactoring, so doing this at same time is maybe too much code change.

NicolasDorier:
supersede by https://github.com/dgarage/NBXplorer/pull/463

PR #445, opened 2023-11-16
Support not creating change with psbt api

Kukks:
This allows you to explicitly specify to donate leftover change to miners.

### DIFF
--- a/NBXplorer/Controllers/MainController.PSBT.cs
+++ b/NBXplorer/Controllers/MainController.PSBT.cs
@@ -323,7 +323,15 @@ namespace NBXplorer.Controllers
 				if (request.SpendAllMatchingOutpoints is true)
 					txBuilder.SendAllRemainingToChange();
 				psbt = txBuilder.BuildPSBT(false);
-				hasChange = psbt.Outputs.Any(o => o.ScriptPubKey == change.ScriptPubKey);
+				hasChange =  psbt.Outputs.Any(o => o.ScriptPubKey == change.ScriptPubKey);
+				if (request.DonateChangeToMiners && hasChange)
+				{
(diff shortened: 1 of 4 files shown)


NBXplorer/Controllers/MainController.PSBT.cs
```diff
@@ -323,7 +323,15 @@ public partial class MainController
...
 					txBuilder.SendAllRemainingToChange();
 				psbt = txBuilder.BuildPSBT(false);
-				hasChange = psbt.Outputs.Any(o => o.ScriptPubKey == change.ScriptPubKey);
+				hasChange =  psbt.Outputs.Any(o => o.ScriptPubKey == change.ScriptPubKey);
+				if (request.DonateChangeToMiners && hasChange)
+				{
+					var txTmp = psbt.GetGlobalTransaction();
+					txTmp.Outputs.RemoveAll(o => o.ScriptPubKey == change.ScriptPubKey);
+					psbt = txBuilder.CreatePSBTFrom(txTmp, false);
+					hasChange =  psbt.Outputs.Any(o => o.ScriptPubKey == change.ScriptPubKey);
```
NicolasDorier, line 332: `hasChange = false` no ?

NicolasDorier:
I am curious, why do you need this?

NBXplorer/Controllers/MainController.PSBT.cs
Kukks: You always have to define a change address to the tx builder if there is not an exact spend, so has change is almost always true unless the requests is specifically crafted. I then remove the change output manually

Kukks:
> I am curious, why do you need this?

It is a request from @silentlink

PR #450, opened 2023-12-05
More flexible wallets API

NicolasDorier:
Rebase of #431

closes https://github.com/dgarage/NBXplorer/issues/429 https://github.com/dgarage/NBXplorer/issues/291

This PR makes use of the enhanced flexivility of the postgres indexer and exposes them over the current API.

Notably it:

* has a new tracked source, Wallet, which can act as a "bag" for individual scripts, or as a parent to other wallets.
* the generate wallet and track wallet now support specifying a parent wallet
* there is an "associate scripts" endpoint which adds specific scripts/addresses to a tracked source to be watched.
* there is an "import coins" endpoint which allows you to tell NBX that a wallet has existing coins that it has not managed to track (as they were possibly created before nbx started tracking the wallet). This endpoint verifies the coins are real by calling gettxout on bitcoin core. If you have a local state of coins, this is a much lighter and scalable version of resotring a wallet's utxos on nbx than the scanutxoset option (which can only run exclusively and takes a lot of resources). This can be improved by also adding a utxoproof that it was mined in a specific block.

@Kukks is the author

---

I extracted part of this PR into https://github.com/dgarage/NBXplorer/pull/453

* For the API is problematic in several way: While the database allow to assign scripts to every tracked source, it might ends up rendering the database in an unusable state.

For example, if a derivation scheme associated a script without properly assigning `descriptors_scripts`, the `CreatePSBT` would create PSBT which can't be signed by the client.

* For the hierarchy, only `WalletTrackedSource` should be able to have children.
* I don't think we should have `ParentWallet` properties spread out in the API, just adding and removing parents should be enough for every case.
* The AssociateScript should only work (as is) with a WalletTrackedSource to avoid having derivation schemes in a weird state.
* More generally, we need an import/export, where you can specify a wallet, then it gives up a data structure which can be restored as-is in another NBXplorer. I don't think we need an API for AssociateScript, this is very low level and prone to fuck up.

A good next step would be an import/export for derivation schemes, extending it later to WalletTrackedSource.


NicolasDorier:
https://github.com/dgarage/NBXplorer/pull/457

PR #457, opened 2024-01-09
Add Group API

NicolasDorier:
Supersede https://github.com/dgarage/NBXplorer/pull/450
This add a new type of tracked source `GroupTrackedSource`.

You can create a group and add other tracked sources as children to it.

Any addresses generated by a child will be added automatically to the group.
Removing a children also remove its addresses from the group.

On top of this, I added an API to add addresses directly to a group via `/cryptos/{cryptoCode}/groups/{groupId}/addresses` API.

Some routes for group API need to be accessed through `/cryptos` and others directly through `/groups`.

Last, I removed abstraction of tracked-sources in the routes of NBXplorer. This is due to the fact that many routes doesn't make sense for all the tracked-sources. (For example CreatePSBT depends on the ability to be able to generate addresses, which a group doesn't have)

### DIFF
--- a/NBXplorer.Client/ExplorerClient.cs
+++ b/NBXplorer.Client/ExplorerClient.cs
@@ -547,6 +547,28 @@ namespace NBXplorer
 			return GenerateWalletAsync(request, cancellationToken).GetAwaiter().GetResult();
 		}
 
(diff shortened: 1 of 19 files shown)


NicolasDorier:
Got ACK from Kukks, we can adjust later if we need more.

PR #479, opened 2024-09-24
[FIX] Ignore check localhost to whitelist on Bitcoin Core

twofaktor:
Local connections (localhost) do not need to be explicitly whitelisted, as connections from 127.0.0.1 are always allowed without restriction by the Bitcoin node.

This PR fixes the code to avoid the localhost is not whitelisted notice on your node


NicolasDorier:
hey @twofaktor sorry for the slow feedback. I couldn't find what you are saying in bitcoin core's source. Are you sure?

twofaktor:
I don't remember if I saw that in the source code or the Bitcoin Core docs. But, it makes sense to me, there is no need for limits or security for localhost network connections. Anyway, I can add a note to inform readers of my guide to ignore that log. I'm closing this PR request. Thanks!

PR #483, opened 2024-11-14
Add Redoc

SuperJMN:
Uses Redoc to generate NBxplorer's documentation by browsing /index.html

### DIFF
--- a/NBXplorer/Program.cs
+++ b/NBXplorer/Program.cs
@@ -1,4 +1,5 @@
-﻿using Microsoft.Extensions.Logging;
+﻿using System.IO;
+using Microsoft.Extensions.Logging;
 using System.Threading.Tasks;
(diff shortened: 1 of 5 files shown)


NicolasDorier:
Doesn't work for me (maybe F5 is fine, but I didn't test this, what matter to me is that it works on published version):


```bash
docker build -t nicolasdorier/nbxplorer:2.5.4 --platform linux/amd64 .
```

Then run this docker compose https://github.com/btcpayserver/btcpayserver/blob/master/BTCPayServer.Tests/docker-compose.yml

With `docker-compose up -d dev`

Browse `http://localhost:32838/` => 404.
`http://localhost:32838/index.html` => 404.


Note that http://localhost:32838/ should be redirected to this file, else nobody will find out that the doc exists.

PR #544, opened 2026-05-26
Add `curl` to Docker image to support health checks

s373nZ:
Adds `curl` as a tool in the Docker image in case operators want to specify HTTP based health checks to ensure HTTP server and/or RPC connectivity is ready before starting dependent containers.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -11,6 +11,8 @@ RUN cd NBXplorer && \
 FROM mcr.microsoft.com/dotnet/aspnet:10.0.6-noble
 WORKDIR /app
 
+RUN apt-get update && apt-get install -y --no-install-recommends curl && rm -rf /var/lib/apt/lists/*
+
 RUN mkdir /datadir
 ENV NBXPLORER_DATADIR=/datadir
 VOLUME /datadir


NicolasDorier:
Sorry I do not understand. How does those health check works? why do you need curl inside the image?

s373nZ:
I could have been more clear about my use case. Let's say I had a Docker compose file like this:
```
  nbxplorer:
    restart: unless-stopped
    expose:
      - "32838"
    healthcheck:
      test: [ "CMD-SHELL", "curl -f http://127.0.0.1:32838/health" ]
      interval: 5s
      timeout: 5s
      retries: 30
    image: nicolasdorier/nbxplorer:2.6.7
```

The command running the health check is running from inside the container, AFAIK. As `curl` and `wget` are not included in the base image `mcr.microsoft.com/dotnet/aspnet:10.0.6-noble`, there isn't an idiomatic way to make HTTP requests to the service to test the NBX HTTP server is up, or that RPCs are being served from `bitcoind`. I looked into trying to make an HTTP request using the dotnet SDK, but didn't work well for me.

Kukks (approved):
ACK, super useful for production

PR #547, opened 2026-08-10
fuzz: add target for FixedSizeCache

brunoerg:
NBXplorer has no fuzz coverage at all so had to write this from scratch. I'm not a C#/.NET expert so did a quick search about fuzz testing tools for it and noticed that SharpFuzz would be the most suitable for it. The idea of this initial target is to exercise the FixedSizeCache with a fuzzable key: a key whose GetHashCode() is directly controlled by the fuzzer.


NicolasDorier:
I appreciate, but I don't want third party libraries for something like this.
No need of a fuzz library for this.

PR #558, opened 2026-09-11
Bound wallet address generation requests

Kukks:
## Dependency

This PR is intentionally stacked on #554 because both changes validate the authenticated derivation/admin request boundary. Review the two-file delta against `feat/nbxplorer-admin-api-validation-pr`; after #554 merges, this branch can be rebased or retargeted to `master` without changing this patch.

## Why

`TrackWalletRequest.DerivationOptions` accepts caller-provided `minAddresses` and `maxAddresses`. Both the synchronous `wait=true` path and the queued `wait=false` path forwarded those integers to address derivation without validating sign, ordering, or size.

`Repository.GenerateAddressesCore` allocates arrays proportional to the requested count, performs CPU-heavy key derivation, and inserts the resulting scripts in 10,000-row chunks. An authenticated client could therefore request an extreme count and create disproportionate memory, CPU, and database work.

## What changed

- Validate every supplied derivation option before any wallet tracking or address generation starts.
- Reject negative `minAddresses` or `maxAddresses`.
- Reject a range where `minAddresses` is greater than `maxAddresses`.
- Reject either value above 10,000, matching the repository's existing maximum insertion chunk.
- Return HTTP 400 with stable code `invalid-address-generation-range` and a concrete constraint message.
- Apply the same validation to both `wait=true` and `wait=false`; the background queue can no longer bypass the bound.

## Preserved behavior

- Null/omitted limits remain valid and retain default gap-pool behavior.
- Zero remains a valid explicit bound.
- Existing bounded requests remain supported; the 500-address end-to-end test passes unchanged.
- Feature selection and address derivation semantics are unchanged after validation.

## Security impact

Per-request address generation is now bounded before proportional allocation, elliptic-curve derivation, or database insertion. This prevents a single authenticated tracking request from scheduling an attacker-selected, effectively unbounded workload.

Addresses Prem Security findings `custos_1t2l0jh` and `custos_hqhxx2`.

## Tests

- Added rejection cases for negative minimum, negative maximum, reversed ranges, and each value above 10,000.
- Added acceptance cases for zero, 500, the exact 10,000 boundary, and omitted values.
- Existing `CanTrackManyAddressesAtOnce` integration test — 1 passed.
- `docker compose build` and `docker compose run --rm tests` from `NBXplorer.Tests` — 109 passed, 0 failed.

## Out of scope

This does not change the configured normal gap-pool size or impose a global rate limit. It constrains only caller-supplied per-request generation counts.

### DIFF
--- a/NBXplorer.Tests/DerivationSchemesControllerTests.cs
+++ b/NBXplorer.Tests/DerivationSchemesControllerTests.cs
@@ -1,0 +1,51 @@
+using NBXplorer.Controllers;
+using NBXplorer.Models;
+using Xunit;
+
+namespace NBXplorer.Tests
+{
+	public class DerivationSchemesControllerTests
+	{
+		[Theory]
+		[InlineData(-1, null)]
+		[InlineData(null, -1)]
+		[InlineData(2, 1)]
+		[InlineData(10_001, null)]
+		[InlineData(null, 10_001)]
+		public void RejectsInvalidAddressGenerationRanges(int? minAddresses, int? maxAddresses)
+		{
+			var request = new TrackWalletRequest
+			{
+				DerivationOptions = new[]
+				{
+					new TrackDerivationOption
+					{
+						MinAddresses = minAddresses,
+						MaxAddresses = maxAddresses
+					}
+				}
+			};
+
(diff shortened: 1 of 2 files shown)


NicolasDorier:
https://github.com/btcpayserver/NBXplorer/pull/560#issuecomment-5662928383

PR #560, opened 2026-09-11
Limit event responses and WebSocket subscription rules

Kukks:
## Why

The event APIs accepted an omitted or arbitrarily large `limit`, which could make a single authenticated request retrieve an unbounded event history. WebSocket clients could also keep appending subscription rules for the lifetime of a connection, making every subsequent event evaluate an attacker-controlled, ever-growing rule list.

NBXplorer is normally deployed as a trusted internal service, so this change focuses on bounding work inside one authorized request or connection. Global concurrent-connection limits remain a deployment/server concern.

## What changed

- Normalize an omitted `GET /events` limit to 10,000 instead of querying an unbounded history.
- Reject event limits below 1 or above 10,000 with the stable `invalid-limit` 400 response.
- Apply the same upper bound to `GET /events/latest`.
- Limit each WebSocket connection to 1,000 total subscription rules.
- Close a socket with `PolicyViolation` before installing an oversized rule set.
- Use overflow-safe accounting when combining existing and newly requested rules.

## Compatibility

Normal polling, long-polling, and WebSocket behavior is unchanged within the limits. Existing clients that omit the event limit still receive up to 10,000 events. WebSocket clients can register up to 1,000 rules per connection.

## Validation

- `dotnet test NBXplorer.Tests/NBXplorer.Tests.csproj -c Release --filter FullyQualifiedName~EventStreamLimitsTests`: 7 passed.
- `dotnet build NBXplorer.sln -c Release --no-restore`: passed.
- `docker compose run --rm tests`: 104 passed, 1 transient cancellation in `CanGenerateWallet` after 69 seconds; the event long-poll and WebSocket integration tests passed.
- Isolated rerun of `CanGenerateWallet`: passed in 40 seconds.
- `git diff --check`: passed.

## Scanner finding

Addresses `custos_h31fyg`: unbounded event history retrieval and per-connection WebSocket subscription growth. This does not add a process-wide connection cap; that remains the responsibility of the trusted deployment boundary and reverse proxy/server configuration.

### DIFF
--- a/NBXplorer.Tests/EventStreamLimitsTests.cs
+++ b/NBXplorer.Tests/EventStreamLimitsTests.cs
@@ -1,0 +1,32 @@
+using NBXplorer.Controllers;
+using NBXplorer.Models;
+using Xunit;
+
+namespace NBXplorer.Tests
+{
+	public class EventStreamLimitsTests
+	{
+		[Fact]
+		public void EventQueryLimitMustBeWithinBounds()
+		{
+			Assert.Equal("invalid-limit", Assert.Throws<NBXplorerException>(() => MainController.ValidateEventLimit(0)).Error.Code);
+			Assert.Equal("invalid-limit", Assert.Throws<NBXplorerException>(() => MainController.ValidateEventLimit(10_001)).Error.Code);
+
+			Assert.Equal(10_000, MainController.NormalizeEventLimit(null));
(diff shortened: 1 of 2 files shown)


NicolasDorier:
NBX isn't meant to be accessible outside, I don't like undocumented limits, crashing is acceptable.

